Issue #115, opened 2014-08-07
Owner and Group Always Set to 0 ("root")


elasticdog:
I'm running confd v0.5.0 on an Ubuntu 14.04 LTS host and am having trouble with altering the owner and group for a template...no matter what I do, it changes things to be owned by the root user and root group, which is not desirable.

**/etc/confd/conf.d/example-json.toml**
```TOML
[template]
src = "example.json.tmpl"
dest = "/etc/example.json"
owner = "elasticdog"
group = "elasticdog"
mode = "0644"
keys = [
  "/jdbc_url",
]
reload_cmd = "/sbin/reload example-server"
```
...and it's creating the file with the desired contents, but the permissions are always set back to 0:

```
$ ls -l /etc/example.json
-rw-r--r-- 1 root root 1011 Aug  7 04:54 /etc/example.json
$ id elasticdog
uid=999(elasticdog) gid=999(elasticdog) groups=999(elasticdog)
$ sudo chmod elasticdog:elasticdog /etc/example.json
$ sudo reload confd
```
**/var/log/upstart/confd.log**
```
2014-08-07T04:54:21Z ip-10-193-17-174 /usr/local/bin/confd[2404]: INFO /etc/example.json has UID 999 should be 0
2014-08-07T04:54:21Z ip-10-193-17-174 /usr/local/bin/confd[2404]: INFO /etc/example.json has GID 999 should be 0
```

popsikle:
+1 to not changing uid:gid unless specified.

crandles:
What about executing `id $username` or `getent group` (if they exist) (`dscacheutil -q group` on osx)? Is it too much a stretch to assume they exist?

kelseyhightower:
We have a solution for this on master. https://github.com/kelseyhightower/confd/commit/6354ddd43eb8589fea16805e935ecdf9491fb7a6

The rules for this are as follows:

If the uid or gid is not set then use the effective uid or gid for the confd process owner. We will no longer default to uid or gid.

zeyuye-airwallex:
does the below work 😄 

```toml
owner = "user"
group = "user"
```

i'm following the docs https://github.com/kelseyhightower/confd/blob/master/docs/quick-start-guide.md#create-template-resources

it uses `mode="nginx"`, however during my experimentation, the destination file permission is root:root

below is my example

`/etc/confd/conf.d/consul.toml`
```bash
[...]

PR #165, opened 2014-10-31
Handle no uid/gid supplied

spheromak:
related to #115 This handles logic for when a uid isn't specified on an existing destination file. The goal is to preserve existing uid/gid on destinations when not specified in the config.

### DIFF
--- a/resource/template/resource_test.go
+++ b/resource/template/resource_test.go
@@ -2,10 +2,13 @@ package template
 
 import (
 	"io/ioutil"
+	"math/rand"
 	"os"
 	"path/filepath"
+	"syscall"
 	"testing"
 	"text/template"
+	"time"
 
 	"github.com/kelseyhightower/confd/backends/env"
 	"github.com/kelseyhightower/confd/log"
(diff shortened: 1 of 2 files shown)


bacongobbler:
If there's someone running tests on unix then we can fix that up in a follow-up, but this LGTM.

@spheromak could you please rebase this on top of master so we can run style tests against this PR?

spheromak:
@bacongobbler rebased

resource/template/resource_test.go
```diff
@@ -40,6 +43,27 @@ keys = [
...
+	ogid := os.Getgid()
+	gid = ogid
+
+	// pick a random group and set the file to that group
+	groups, err := os.Getgroups()
+	if err != nil {
+		return
+	}
+
+	// It is possible this never terminates assert the length is greater than 1
```
bacongobbler: This logic is a bit concerning. I think it'd be better to use a concrete group ID in the tests so we don't have to do funky branch logic like this.

kelseyhightower: +1

spheromak: how do you suppose we do that ? non root group that is always on any possible system ?

kelseyhightower:
We have a solution for this on master. https://github.com/kelseyhightower/confd/commit/6354ddd43eb8589fea16805e935ecdf9491fb7a6

The rules for this are as follows:

If the uid or gid is not set then use the effective uid or gid for the confd process owner. We will no longer default to uid or gid.

PR #179, opened 2014-12-15
Add simple zookeeper support

sheepkiller:
* watch is not yet supported - see backends/zookeeper/client.go for details
  * integration should work
  * add a simple example in quickstart guide
  * exit if watch is enabled when zookeeper is defined as backend

Tested with go 1.4


thomascramer:
I've been waiting for this!  When can this get merged in?

PR #198, opened 2015-02-10
bump go-etcd to v0.4.6; add Makefile targets for building from source

### LINKED ISSUE #195: latest etcd client tries to load /v2/members rather than /v2/machines
Running a new build of confd against an existing etcd 0.4.6 it fails to identify the cluster even though it is reachable with curl. confd returns the error

    ERROR cannot connect to etcd cluster: 172.17.8.102:4001

But `curl http://172.17.8.102:4001/v2/machines` succeeds.

I pulled down the confd code and traced through it to discover what was happening. It looks like the etcd client now looks for /v2/members as of 3 days ago - [see this commit](https://github.com/coreos/go-etcd/commit/5348c80bde6223fa53c48ef546b4802db69b1d70#diff-33c1223aac4ef741731cc8c73b4d7777L295)

Of course, when run against an older etcd, this fails and produces a 404, and confd fails.

I think this is a robustness problem for etcd rather than confd, but I thought it worth logging here in case others start experiencing it.

bacongobbler:
This PR is two-fold: one commit is to close #195 by switching go-etcd to v0.4.6 for etcd clusters which are pre-2.0.

Adding the Makefile + docs is another response to #195. Assuming that users were running `godep go build` to build confd, they should not have run into this issue since coreos/go-etcd@23142f6 should have worked fine for most users. That commit dates back before the breaking change. Having a unified tool to help build/install the confd binary should prevent these issues in the future. :)

closes #195


kelseyhightower:
@bacongobbler Looks like we need to rebase this.

johanneswuerbach:
The zk dependency was actually not vendored, I added zk to Godeps here: https://github.com/kelseyhightower/confd/pull/206

bacongobbler:
nice catch. :+1:

bacongobbler:
gentle ping

bacongobbler:
@kelseyhightower if you're too busy to maintain the project, I'd be more than happy to help out :)

bacongobbler:
okay, rebased again. Go travis go

kelseyhightower:
@bacongobbler This got pushed to master: https://github.com/kelseyhightower/confd/releases/tag/v0.8.0

bacongobbler:
awesome. So now the only thing that's missing is the documentation around using the Makefile so users stop `go get`'ing confd... If they do, they'll get go-etcd from master :(

I'll amend this so we can get that sucker in

Issue #183, opened 2015-01-04
confd zero-conf mode


rdsubhas:
Hi, Firstly thanks for the awesome library. The only problem we're facing is, confd itself requires a bunch of configuration files. We're having a bunch of services, each having its own service configuration as a template. But just for confd to play well, we're having to create `/etc/confd/...` templates. It also means that since the paths are dynamic, we can't hardcode paths in the TOML template as well, and we're having to manually construct a TOML file in code.

Is there any possibility of confd running in a zero-configuration mode? i.e. Just using command line arguments, I specify everything, including the source template, target file, backend and everything else? That would be really awesome!

kelseyhightower:
@rdsubhas +1 I think the next release of confd should support this and it should go in before 1.0.0.

abelaska:
+1

rdsubhas:
:+1: @kelseyhightower Thanks so much!

diegomarangoni:
+1

nikfoundas:
@rdsubhas I had a similar idea to include all configuration files and templates in the underlying cluster registry itself (for example inside etcd) using the same structure as in the filesystem. This would speed up deployment in container based environments like coreos docker containers. Most probably one more flag would be required in order to specify whether configuration should be read from file system or the registry.

@kelseyhightower Very nice piece of software!

mishak87:
:+1:

kelseyhightower:
I'll be hacking on this today. Will provide updates soon.

johnrengelman:
I've been considering this a lot lately. I've been using Rancher heavily and it's nice that confd supports the Rancher Metadata API backend natively. However, I have the same issue. Everytime I want to provide a configuration file to an app, I need to write a custom container that contains `confd` and the necessary configuration.
What I would really like is if `confd` could either use CLI args or the Rancher metadata API to provide the configuration.

I think @nikfoundas's idea of having a `config-backend` would be awesome. The default would be `filesystem`, so it would be something like:
`--config-backend=filesystem --config-option=dir=/etc/confd.d`,
But we could also have something like:
`--config-backend=rancher --config-option=prefix=/2015-07-25 --config-option=key=/self/service/metadata/confd`

I personally like the `--config-option` flag because it would allow additional backends without creating a large assortment of CLI args to deal with, though I could go either way.

An outstanding question here is if `confd` would "watch" the config backend for changes, or if it would be a one-time shot at startup (I think this would be sufficient)

[...]

PR #201, opened 2015-02-12
add filesystem support

glerchundi:
* watch is not yet supported - see backends/fs/client.go for details
* integration should work
* exit if watch is enabled when fs is defined as backend

Tested with go 1.4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,4 +29,5 @@ script:
   - ./test
   - bash integration/consul/test.sh
   - bash integration/etcd/test.sh
+  - bash integration/fs/test.sh
   - bash integration/redis/test.sh
--- a/docs/quick-start-guide.md
+++ b/docs/quick-start-guide.md
@@ -12,6 +12,7 @@ confd supports the following backends:
 * redis
 * zookeeper
 * dynamodb
+* filesystem
 
 ### Add keys
 
@@ -74,6 +75,16 @@ aws dynamodb put-item --table-name <YOUR_TABLE> --region <YOUR_REGION> \
 aws dynamodb put-item --table-name <YOUR_TABLE> --region <YOUR_REGION> \
     --item '{ "key": { "S": "/myapp/database/user" }, "value": {"S": "rob"}}'
 ```
+
+#### filesystem
+
+```
+mkdir -p /folder/myapp/database
+printf "db.example.com" > /folder/myapp/database/url
+printf "rob"            > /folder/myapp/database/user
+```
+
+`-fs-rootpath` is available to provide a custom root path, in the example above it would be used like this `-fs-rootpath="/folder"`.
 
 ### Create the confdir
 
--- a/integration/fs/test.sh
+++ b/integration/fs/test.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# get full path: http://stackoverflow.com/questions/4774054/reliable-way-for-a-bash-script-to-get-the-full-path-to-itself
+pushd `dirname $0` > /dev/null
+ROOTPATH=`pwd -P`
+popd > /dev/null
+
+mkdir -p $ROOTPATH/database
+mkdir -p $ROOTPATH/upstream
+mkdir -p $ROOTPATH/prefix/database
+mkdir -p $ROOTPATH/prefix/upstream
+
+printf "foobar"         > $ROOTPATH/key
+printf "127.0.0.1"      > $ROOTPATH/database/host
+printf "p@sSw0rd"       > $ROOTPATH/database/password
+printf "3306"           > $ROOTPATH/database/port
+printf "confd"          > $ROOTPATH/database/username
+printf "10.0.1.10:8080" > $ROOTPATH/upstream/app1
+printf "10.0.1.11:8080" > $ROOTPATH/upstream/app2
+printf "127.0.0.1"      > $ROOTPATH/prefix/database/host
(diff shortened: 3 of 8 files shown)


backends/fs/client.go
```diff
@@ -0,0 +1,63 @@
...
+	"strings"
+)
+
+type Client struct {
+	rootPath    string
+	maxFileSize int64
+}
+
+func NewFsClient() (*Client, error) {
+	// TODO: how to provide rootPath in order to pass tests?
```
bacongobbler: You can use `os.Setenv()` in the tests

backends/fs/client.go
```diff
@@ -0,0 +1,63 @@
...
+func (c *Client) GetValues(keys []string) (map[string]string, error) {
+	vars := make(map[string]string)
+	for _, key := range keys {
+		keyPath := path.Join(c.rootPath, key)
+		if _, err := os.Stat(keyPath); os.IsNotExist(err) {
+			return vars, err
+		}
+
+		walk := func(fpath string, fi os.FileInfo, err error) error {
+			if (err != nil) {
```
bacongobbler: From https://golang.org/doc/effective_go.html#control-structures, parentheses in control structures are not allowed. You should've seen compile errors IIRC

backends/fs/client.go
glerchundi: yeah, but os.Getenv(...) probably isn't the way to go. Maybe custom params for concrete backends in command line arguments? I put there just as a reminder, it's already being used in the `integration/test.sh` to get current testing folder.

bacongobbler: You could just pass in rootpath in the signature.

```
func NewFsClient(rootPath string, maxFileSize int64) (*Client, error) {
    ...
}
```

glerchundi: The idea was to provide a simple manner to use the existing integration tests and make them work with this new backend, nothing more. 'rootPath' could be achieved with prefix anyway.

glerchundi:
Hi,

@bacongobbler @kelseyhightower 
Any update on this?

I'm going to explain why I found this useful. And I would like to know if there is a possibility to incorporate this on the main branch. If not, please tell me something!

I would like to keep secrets, certificates and everything related to security in etcd and then be able to use them inside a docker container which will be launched kubernetes:
- create a kubernetes volume, EmptyDir (from now on /volatile-vol/) one which is volatile by default (will be deleted when pod is destroyed).
- dump from a kvstore (etcd, zookeeper, whatever) to the `/volatile-vol`. For example, dump from `/secrets/databases/canary/db_password` to `/volatile-vol/DB_PASSWORD`
- use confd to create our configuration files based in `/volatile-vol/DB_PASSWORD`

Benefits:
- filter data from the backend and make available to the container only what matters to it.
- separation of concerns, I would like to hide to the container create how is structured our backend, just follow our patterns which define secrets would be available in `/volumes/secrets`
- there are already utilities which load environment variables from an envdir, for example, `s6-envdir`. We use it in this way: [`with-contenv`](https://github.com/just-containers/s6-overlay-builder/blob/v1.8.2/builder/overlay-rootfs/usr/bin/with-contenv). Dumping kvstore keys into directory facilitates patterns like this.

What do you think?

bacongobbler:
Hmm, couple of thoughts here:

 - there should be a way for users to configure the max file size.
 - I know that there's interest in #183, so the maxFileSize and rootPath should be part of the backend's config for consistency, similar to https://github.com/kelseyhightower/confd/blob/master/backends/client.go#L34.
 - `--watch` should be [disabled entirely](https://github.com/kelseyhightower/confd/blob/master/config.go#L151-L154).

You'll also want to rebase off master so this PR can take advantage of the new test infra: https://github.com/kelseyhightower/confd/commit/1bc750778236c0bd4de55ab59bec600f754456b8

thanks!

glerchundi:
Ok, I'll do it during this week,

Thanks!

glerchundi:
Done.

- `fs` is included in the unsupported backends list.
- fs integration tests were included in the travis test infra.
- I really dislike the idea of mixing global opts with custom backend opts but, right now and without doing an important refactoring, I don't see how could be this done elegantly. So for now, I exposed `-fs-rootpath` and `-fs-maxfilesize`. A possible solution for the future could involve using `codegansta/cli` to create something like `confd [global opts] etcd/consul/fs/whatever [backend opts]`. If you guys @bacongobbler and @kelseyhightower think this is an interesting behaviour feature I could give it a try.

bacongobbler:
> A possible solution for the future could involve using codegansta/cli to create something like confd [global opts] etcd/consul/fs/whatever [backend opts]. If you guys @bacongobbler and @kelseyhightower think this is an interesting behaviour feature I could give it a try.

I personally prefer to stick with the native flag package for handling option flags. That way we don't have to rewrite the entire CLI just to wrap around an opinionated lib. The standard library is pretty slick. :)

Regardless, I agree that there should be a better way to set up a backend's config options rather than through global flags. I'll keep thinking about how we can roll this out, but so far this implementation is about as clean as it gets without going too crazy.

kelseyhightower:
@glerchundi Rebase please. I'm reviewing all the PRs today.

glerchundi:
@kelseyhightower, done. travis is not detecting changes on PR, is it possible to force a new one?

scalp42:
@glerchundi any chance you could rebase against master please?

@kelseyhightower any round of PR reviews coming soon please?

The use case for this backend is :+1:

glerchundi:
@scalp42 done :-)

PR #203, opened 2015-02-14
Fmt backends zookeeper client.go

cinience:


bacongobbler:
Thanks! We should probably do some [gofmt validation](https://github.com/fishworks/dis/blob/master/Makefile#L43-L46) on PRs so this doesn't happen again.

cinience:
that's awesome!

PR #206, opened 2015-02-23
Added missing zk godep

johanneswuerbach:
The zk godep is missing and CI is currently only passing, because Travis runs `go get` by default, which downloads zk https://travis-ci.org/kelseyhightower/confd/jobs/51791525#L110


bacongobbler:
LGTM!

PR #235, opened 2015-03-24
log that we're skipping config file to debug

bacongobbler:
In recent versions of confd, this file is not necessary to run confd
as seen in the integration tests. This switches the log to debug as it's
still interesting information, but it's not necessary to log if we've
supplied the correct flagset to run confd without a config file.

### DIFF
--- a/config.go
+++ b/config.go
@@ -104,7 +104,7 @@ func initConfig() error {
 	}
 	// Update config from the TOML configuration file.
 	if configFile == "" {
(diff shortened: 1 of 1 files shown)


mboersma:
That would be preferable to always seeing this kind of log line:
```
2015-03-24T17:09:37Z 97df176fe07e confd[81]: WARNING Skipping confd config file.
```
+1.

scalp42:
Any chance to see this merged please ?

PR #238, opened 2015-03-25
client cert config from environment variables

alexnauda:
This allows confd to take cert-related configuration from environment variables. Could be expanded to handle all of the configuration variables.

### DIFF
--- a/config.go
+++ b/config.go
@@ -85,7 +85,8 @@ func init() {
 }
 
 // initConfig initializes the confd configuration by first setting defaults,
-// then overriding setting from the confd config file, and finally overriding
+// then overriding settings from the confd config file, then overriding
+// settings from environment variables, and finally overriding
 // settings from flags set on the command line.
 // It returns an error if any.
 func initConfig() error {
@@ -116,6 +117,10 @@ func initConfig() error {
 			return err
 		}
 	}
+
+	// Update config from environment variables.
+	processEnv()
+
 	// Update config from commandline flags.
 	processFlags()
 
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -116,6 +117,10 @@ func initConfig() error {
 			return err
 		}
 	}
+
+	// Update config from environment variables.
+	processEnv()
```
bacongobbler: I think this should go before we load the config file.

bacongobbler:
small nit, but otherwise LGTM!

config.go
alexnauda: Are you suggesting that the config file should override the environment variables?

mboersma:
+1. Should there be a mention of `CONFD_CLIENT_CAKEYS` and friends in [configuration-guide.md](https://github.com/kelseyhightower/confd/blob/master/docs/configuration-guide.md)?

timfallmk:
:+1: @bacongobbler Agreed that config files should override the env variables, if present.

config.go
bacongobbler: yes indeed

alexnauda:
Agreed on the documentation for this feature... Please don't merge it until I get that in there.

As for the precedence of configuration file over environment variables, I can't think of another utility that behaves that way. Take for example etcd, for which "Options set on the command line take precedence over all other sources. Options set in environment variables take precedence over options set in configuration files."
https://coreos.com/docs/distributed-configuration/etcd-configuration/

timfallmk:
I agree that command line flags should override configuration files, but they're set at run time so it's obvious they're being called. Environment variables can be set and forgotten, so it might be confusing if you're changing config settings and they're not being applied.Perhaps the solution is a warning. Something along the lines of "blah blah setting take from environment variable".Sent using CloudMagicOn Sat, Apr 4, 2015 at 01:03,  kelseyhightower/confd <reply@reply.github.com> wrote:            Agreed on the documentation for this feature... Please don't merge it until I get that in there.

As for the precedence of configuration file over environment variables, I can't think of another utility that behaves that way. Take for example etcd, for which "Options set on the command line take precedence over all other sources. Options set in environment variables take precedence over options set in configuration files."
https://coreos.com/docs/distributed-configuration/etcd-configuration/

bacongobbler:
I'm still adamant about config files taking preference over envvars due to the human error of "set and forget" which @timfallmk mentioned, especially when running two separate `confd` processes on the same host. Typically you'll only run a single `etcd` instance on a machine so it doesn't really affect them all that much, but it can be a common problem with confd.

@kelseyhightower, any input?

kelseyhightower:
This one is tricky. I think it's OK if we go with flag > env var > config file.

kelseyhightower:
But... The config file today where all the action happens. What do others think on this one?

bacongobbler:
flag > envvar > config file is fine for me, though a future enhancement should be that we log where the config is being retrieved from as a DEBUG message. That way there are no surprises (at least when someone tries to use `--debug`)

PR #245, opened 2015-03-31
Enable watch from the config file

### LINKED ISSUE #208: Setting watch to true only works from the command-line
Setting watch=true in /etc/confd/confd.toml has no effect, even if it is documented that it should.

A quick look at confd.go (pardon my weak go-foo) seems to indicate that to me. It's picking the processor based on cmdline arguments early and doesn't look at config.Watch

        switch {
        case watch:
                processor = template.WatchProcessor(templateConfig, stopChan, doneChan, errChan)
        default:
                processor = template.IntervalProcessor(templateConfig, stopChan, doneChan, errChan, config.Interval)
        }

bookest:
Support enabling watch from either the command line or the config file.
 
This fixes #208.

### DIFF
--- a/confd.go
+++ b/confd.go
@@ -38,7 +38,7 @@ func main() {
 	errChan := make(chan error, 10)
 	var processor template.Processor
 	switch {
-	case watch:
+	case config.Watch:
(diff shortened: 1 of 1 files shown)


bacongobbler:
LGTM

PR #175, opened 2014-12-04
Initial conversion to a package

bakins:
Very "brute-force" attempt.  Bare minimum to convert to a package.  Avoided rearranging any code.  Runs and passes `go test`, but not much testing beyond that.  Consider this a conversation starter.

There were a few things I would like to do along with this, but I wanted to avoid doing too much in an initial attempt.

PR #265, opened 2015-04-11
refactor to packetize and a whole mess of things

glerchundi:
Hi,

This is going to be a good one!

I made a confd refactoring, changes include primarily how input parameters are treated and how to define cli arguments. Everything is configured via config structs (it helps packetizing confd #175) and cli arguments are created dynamically introspecting StructTags. Take a look to `config/backend.go`, `config/global.go`, and `config/template.go`. when it comes to cli handling `codegangsta/cli` library was used.

The most important thing in these changes is that I always follow the same pattern:
 - load from configuration file
   - for global config options [global] section in confd.toml is used.
   - for backend config options [etcd/consul/fs/...] section in confd.toml is used.
   - for template config options [template] is used in each template-defining file.
 - override if envar is defined and has a value
 - override if cli arg is provided

So the precedence order is as follows:
 - cli options
 - env var
 - configuration file

Also, I implemented the zero-conf mode (#183) using `-tmpl-*` cli arguments. Those are the same as the ones used in the configuration files. It can be rendered just one template per confd execution. Note that if confdir is provided zero-conf mode is ignored. It can be used like this

```
KEY=myvalue1 confd --onetime --tmpl-src "test.txt.tmpl" --tmpl-dest "test.txt" env
```

Yes, it is really opinionated but I think it is somehow organized and helps more than what it distrubs :-).

What do you think? @kelseyhightower @bacongobbler I would like to know your opinion on this, thanks!

PD.: This gist contains all configuration file examples and cli help outputs: https://gist.github.com/glerchundi/9b0fe054a157bc0eff24.


kelseyhightower:
@glerchundi This is a huge PR, and will take a lot of time to review. It might turn out to be the case that there is to much refactoring here and I may ask that we break this up. I need to review first, but I just wanted to give you the heads up.

glerchundi:
cool! if you just want to try it out, I compiled confd as you usually do:

https://github.com/glerchundi/confd/releases/download/v0.10.0-beta1/confd-0.10.0-beta1-darwin-amd64
https://github.com/glerchundi/confd/releases/download/v0.10.0-beta1/confd-0.10.0-beta1-linux-amd64

glerchundi:
me again, I've created a test container which uses this confd version to load-balance http requests using nginx:

Concretely these are the files using confd:

https://github.com/just-containers/nginx-loadbalancer/blob/master/rootfs/etc/cont-init.d/01-confd-onetime
https://github.com/just-containers/nginx-loadbalancer/blob/master/rootfs/etc/services.d/confd/run

kelseyhightower:
After a deeper review I'm not ready to move to a new package for managing flags. The goal right now is to clean up the backends, and a few more tests and get 1.0.0 out the door.

I think we can revisit this later, but I don't want to introduce such a big overhaul at this stage.

PR #271, opened 2015-04-18
feat: add strings.Contains support

### LINKED ISSUE #270: add strings.Contains support
related: #252 

We want to check if a certain domain is a subdomain of another route in our nginx template so we can apply similar properties to the specific route.

carmstrong:
closes #270


kelseyhightower:
LGTM

PR #274, opened 2015-04-20
Draft: Support for DynamoDB

bracki:
This pull request adds support for DynamoDB.

You can use it via `confd -backend dynamodb -table my_table`.
It assumes you have previously created the table `my_table` in DynamoDB.
All AWS related credentials can be setup via all the mechanisms that the Go AWS SDK [supports](https://github.com/awslabs/aws-sdk-go/blob/master/aws/credentials.go#L46).

The schema used for the table can be found [here](https://github.com/bracki/confd/commit/9fd122d0acc77c59e2770c1ab6e06e0a6626ceb2).


bracki:
Ok, what currently breaks is the Go 1.2 tests because the AWS SDK uses the `Timeout` field on the `http.Client`.
All AWS settings can be configured either via environment variables or EC2 specific instance profiles.
The only thing missing is to make the table configurable. Would like some input on that. Should I just stick it into `Config`?

kelseyhightower:
@bracki I think we should be able to move past go 1.2 at this point.

kelseyhightower:
@bracki I've merged this but we may need to clean it up a bit before the next confd release.

bracki:
@kelseyhightower Cool! What exactly do you think needs to be cleaned up? Just let me know and I'll take care. BTW we are using this in production on a 30+ instance auto scaling group.

PR #279, opened 2015-04-26
allow ./build to be run directly from downloaded source packages...

schlitzered:
...without cloning the code to $GOPATH/src.

this is done the same way as it is done in the etcd project.

### DIFF
--- a/build
+++ b/build
@@ -7,6 +7,12 @@ GO="godep go"
 ORG_PATH="github.com/kelseyhightower"
 REPO_PATH="${ORG_PATH}/confd"
 
+export GOPATH=${PWD}/gopath
+
+rm -f $GOPATH/src/${REPO_PATH}
+mkdir -p $GOPATH/src/${ORG_PATH}
+ln -s ${PWD} $GOPATH/src/${REPO_PATH}
+
 echo "Building confd..."
 # Static compilation is useful when confd is run in a container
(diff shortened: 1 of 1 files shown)


build
```diff
@@ -7,6 +7,12 @@ GO="godep go"
 ORG_PATH="github.com/kelseyhightower"
 REPO_PATH="${ORG_PATH}/confd"
 
+export GOPATH=${PWD}/gopath
+
+rm -f $GOPATH/src/${REPO_PATH}
```
bacongobbler: This is concerning if you're running this script via $HOME and you have your normal $GOPATH set to `gopath`:

```
$ pwd
/Users/bacongobbler
$ ls | grep gopath
gopath
$ ./gopath/src/github.com/kelseyhightower/confd/build
```

Which will delete your confd changes.

schlitzered: you are right. line 10 should be replaces with something like this:

SCRIPT=`realpath $0`
SCRIPTPATH=`dirname $SCRIPT`

export GOPATH=$SCRIPTPATH/gopath


what do you think?

bacongobbler:
Sounds good to me! I've used the following in other projects as a shortcut: `ROOT_PATH=$(cd $(dirname $0); pwd)`

schlitzered:
i guess using dirname $0 could also break thinks, as it should point to the wrong directory if the script is a symlink?

i will create a new pull request.

PR #281, opened 2015-04-27
allow ./build to be run directly from downloaded source packages...

schlitzered:
finds the realpath of the build script, and sets the gopath to the local source


kelseyhightower:
We are now using the [gb tool](http://getgb.io/) to manage builds and dependencies.

PR #291, opened 2015-05-13
Log command output on error

philk:
We want to see the command output whether the command exited in error or
not. Otherwise it's very difficult to debug why your command is failing.


bacongobbler:
@philk I think the better change would be to move the debug log lines on 217 and 231 to before the error check. That way you can turn on `--log-level debug` and check the output. That was an oversight from my PR (commands not being displayed when an error occurs), sorry!

philk:
@bacongobbler yeah, I actually did that originally but I think most users would want the command output in the case of an error even if they don't have debug logging on. What do you think?

bacongobbler:
Good point. I'm happy with the implementation. LGTM!

kelseyhightower:
LGTM

PR #294, opened 2015-05-25
Use the gb tool for dep management and building.

kelseyhightower:
Currently confd uses the go tool for running tests and building.
Godeps is used to manage dependencies.

This commit moves the project to using the [gb tool](http://getgb.io/)
for running tests, building, and managing dependencies.

The build, tests, and travis files have been updated to use gb.


kelseyhightower:
@bacongobbler I would love to get your thoughts on this before I make this change. I would live to move to using the gb tool to manage confd going forward. This is mainly because Godep is really hard to work with when adding new dependancies.

I also like the new workspace workflow offered by gb.

bacongobbler:
LGTM

PR #297, opened 2015-05-27
Use prefix instead of Prefix when monitoring changes.

Who828:
We never populate Prefix when creating a new templateResource, hence it's always empty when setting up a watch. I ran into this when I was trying to add support for watches with ZK backend.

I am not sure why we need Prefix in template to begin with (also it's confusing to have Prefix and prefix in the same struct), if you want I can refactor Prefix across the codebase.


kelseyhightower:
@Who828 ping.

bacongobbler:
perhaps adding a test would help me be more confident in this fix, as I'm not sure what this would affect within the codebase. It might be better to refactor it across the codebase.

kelseyhightower:
This has been fixed on master:
https://github.com/kelseyhightower/confd/commit/30663b9822fe8e800d1f2ea78447fba0ebce8f6c

PR #210, opened 2015-02-28
Template integration with https://github.com/xordataexchange/crypt provided values

philipsoutham:
Template integration with https://github.com/xordataexchange/crypt provided values

New cli option `-secret-keyring` -> path to armored secret keyring

| template function | crypt enabled function |
| --- | --- |
| `get` | `cget` |
| `gets` | `cgets` |
| `getv` | `cgetv` |
| `getvs` | `cgetvs` |

Why? Because I wanted to store SSL certs along with the nginx config values.

Also, sorry; I noticed my godeps got attached to the PR.

Closes #210 #342

PR #300, opened 2015-06-04
New backend, etcd+crypt

jakexks:
After stumbling across https://github.com/xordataexchange/crypt we thought we should use it, and therefore would need to modify confd to include it!


jakexks:
Build fails due to the golang crypto library in vendor/src being a subproject from https://github.com/golang/crypto/, wasn't sure whether to include it in the commits or not!

bacongobbler:
isn't this a duplicate of #210?

jakexks:
Ha, didn't notice someone had already done it. It's a slightly different approach though, this blanket decrypts everything on the way out of etcd when you chose the backend "etcdcrypt" rather than adding new template functions to deal with encrypted values, meaning you can use your existing templates unmodified.

bacongobbler:
Personally I'm leaning towards the latter's approach. Working with etcd secrets is probably an edge case for a few certain keys. After that I'd want access to unencrypted values. That's my opinion though. @kelseyhightower?

jakexks:
I agree the method in #210 is definitely more flexible. Our use case is multiple parties configuration being stored in the same etcd cluster, meaning we would blanket encrypt / decrypt with a unique key per party - so it never occurred to me to extend the template parser.

kelseyhightower:
I'm closing this in favor of #210

PR #314, opened 2015-07-27
Proposal feature: Allow templates to be sourced from backend

crandles:
Allow confd to source the templates from the configured backend via a provided key.

Example:
* /etc/confd/conf.d/myconfig.toml
```
[template]
src_key = "/templates/myserviceconfig"
dest = "/etc/my/config.conf"
keys = [
    "/myapp/database/url",
    "/myapp/database/user",
]
```
* /templates/myserviceconfig key contents:
```
# confd generated config file
# from remote template
database_url = {{getv "/myapp/database/url"}}
database_user = {{getv "/myapp/database/user"}}
```

There are some different issues filed for creating dynamic configs, and I think this would help move towards supporting that. For my use case, I'd like to be able to update the sourced templates in `etcd` and have them propagate automatically. Without something like this I may have to orchestrate updating templates on a large number of servers.

### DIFF
--- a/src/github.com/kelseyhightower/confd/resource/template/resource.go
+++ b/src/github.com/kelseyhightower/confd/resource/template/resource.go
@@ -44,6 +44,7 @@ type TemplateResource struct {
 	Mode          string
 	Prefix        string
 	ReloadCmd     string `toml:"reload_cmd"`
+	SrcKey        string `toml:"src_key"`
 	Src           string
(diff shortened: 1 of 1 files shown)


kelseyhightower:
I'm going to revisit the idea of fetching templates from a remote source. I think we need to support http and the supported backends. I'm closing this for now, but I'll keep the issue open.

Issue #310, opened 2015-07-10
Add support for dynamic keys


diegomarangoni:
This was mentioned before by @duffqiu in [this issue](https://github.com/kelseyhightower/confd/issues/73).

Would be nice to prioritize the services running in the same machine as my nginx with different weight.
But, for that works, I need to watch a key in a dynamic path that contains a dynamic token, like the machine id.

```
[template]
src = "nginx.tmpl"
dest = "/etc/nginx/conf.d/example.com"
keys = [ "/announce/services/hhvm", "/announce/machines/${CURRENT_MACHINE_ID}/services/hhvm" ]
owner = "nginx"
mode = "0644"
reload_cmd = "service nginx reload"
```

And at the template, loop the both keys, but excluding elements of first loop that are in second one, like that:

```
upstream hhvm_backend {
    least_conn;

    server 127.0.0.1 down;

{{ range getvs "/announce/machines/${CURRENT_MACHINE_ID}/services/hhvm/*" }}
    server {{ . }}:9000 weight=5;
{{ end }}

{{ range getvs "/announce/services/hhvm/*" }}
    ### at this loop, exclude the elements of the previous one ###
    server {{ . }}:9000;
{{ end }}

    keepalive 8;
}

server {
    listen 80 default_server;

    server_name .example.com;
    root /var/www/example.com;

    location / {
        try_files $uri /index.php$is_args$args;
    }

    location ~ \.(hh|php)$ {
        include fastcgi_params;

        fastcgi_param     REQUEST_METHOD $request_method;
        fastcgi_param     SCRIPT_FILENAME $document_root$fastcgi_script_name;
        fastcgi_index     index.php;
        fastcgi_pass      hhvm_backend;
        fastcgi_keep_conn on;
    }
}
```
This could be achieved by doing some kind of named keys, like:

```
[template]
keys = {
    "all" => "/announce/services/hhvm",
    "local" => "/announce/machines/${CURRENT_MACHINE_ID}/services/hhvm"
}
```
and then:
```
upstream hhvm_backend {
    least_conn;

    server 127.0.0.1 down;

{{ range keys("local") }}
    server {{ .Value }}:9000 weight=5;
{{ end }}

{{ range keys("all") }}
{{ if nin(.Key, keys("local")) }}
    server {{ .Value }}:9000;
{{ end }}
{{ end }}

    keepalive 8;
}
```

[...]

Issue #256, opened 2015-04-09
Create multiple config files from templates


lorenz:
I want to run tinc with confd, tinc requires you to have a configuration file for every host. These can all be built from the same config. I'd like to have the host list and the host-specific configuration stored in etcd and afterwards turned into files by confd. Any chance that this is getting added?

kelseyhightower:
@lorenz What would that look like? What data would each template receive?

lorenz:
@kelseyhightower Basically there would be a go template which is the same for every file. The data would be equal (structure-wise, not data-wise) but grouped up in directories, one for every file. Does that make sense?

mborho:
+ 1

The possibility to generate config files programmatically would be very handy.

kelseyhightower:
@mborho Can you give me an idea of what an example resource template and template file would look like?

mborho:
@kelseyhightower Good question :)
Let me start with the use case first. Having some keys in the following way:
```
/auth/foo/user1
/auth/foo/user2
/auth/bla/user1
```
Which should result  in multiple files after processing:
```
foo.htpasswd
bla.htpasswd
```
Currently I'm using an intermediate file, which gets cut by some awk magic added to the *restart_cmd*.

But perhaps a custom tag is possible in the templates, something like this:
```
{{configfile "filename" "fileprefix" "content"}}
```
And in the template resource something like:
```
dest_dir_multiple = /path/for/configfiles/
```
So the config-files would be saved as */path/for/configfiles/[filename].[fileprefix]*

Makes this any sense for you?

crunchywelch:
FWIW, this maybe addresses this issue: https://github.com/kelseyhightower/confd/pull/332

kelseyhightower:
After thinking about this for a while, I'm going to reject this feature in order to keep confd simple. confd is really designed to map one template resource to a single output file. This keeps confd simple, but as you pointed out it does limit the types of problems confd can solve.

I'm willing to make this trade off though in order to keep confd simple.

openxxs:
Hi, @bacongobbler I need to confd a nginx.conf file and a ./conf.d/upstream.conf file, which nginx.conf includes ./conf.d/upstream.conf. Because these two files are determined by different etcd clusters, I have to start a confd process for each.
toml file for nginx.conf:
```
[template]
prefix = "/domeos/loadbalancer/nginx/nginx03"
src = "nginx.conf.tmpl"
dest = "/etc/nginx/nginx.conf"
[...]

PR #315, opened 2015-07-27
Proposal feature: Support multiple template resources per file

crandles:
This will break the template config syntax, so it may be necessary to alter in some way to provide backwards compatibility. That being said, this PR would allow for a single Template Resource Config file to contain multiple template definitions.

* Use case(s):
This would allow for the dynamic generation of confd templates in the following manner
**Example**: Initial confd template resouce is defined, which then dynamically creates a second confd resource template based on data retrieved from backend.

**/etc/confd/conf.d/default.toml**
```
[[template]]
src = "nginx.toml.tmpl"
dest = "/etc/confd/conf.d/nginx.toml"
keys = [
    "/nginx/",
]
```
**/etc/confd/templates/nginx.toml.tmpl**
```
# In this example, every domain configured under "/nginx/*" gets its own config file.
{{range gets "/nginx/*}}
{{$data := json .Value}}
[[template]]
src = "nginx.conf.tmpl"
dest = "/etc/nginx/conf.d/{{$data.domain}}.conf"
keys = [
    "/nginx/{{$data.domain}}/",
    "/other/path/{{$data.other}}/",
]
{{end}}
```
I've left out the final `nginx.conf.tmpl`, that would be a standard confd template used to generate a single `nginx/conf.d/$domain.conf` file in this example.

I think this could solve for #310 and #256 (and other dynamic-config generation use cases) in a clean manner.

### DIFF
--- a/src/github.com/kelseyhightower/confd/resource/template/processor.go
+++ b/src/github.com/kelseyhightower/confd/resource/template/processor.go
@@ -124,7 +124,7 @@ func getTemplateResources(config Config) ([]*TemplateResource, error) {
 			lastError = err
 			continue
 		}
-		templates = append(templates, t)
+		templates = append(templates, t...)
 	}
 	return templates, lastError
 }
--- a/src/github.com/kelseyhightower/confd/resource/template/resource.go
+++ b/src/github.com/kelseyhightower/confd/resource/template/resource.go
@@ -31,7 +31,7 @@ type Config struct {
 
 // TemplateResourceConfig holds the parsed template resource.
 type TemplateResourceConfig struct {
-	TemplateResources []TemplateResource `toml:"template"`
+	TemplateResources []*TemplateResource `toml:"template"`
 }
 
 // TemplateResource is the representation of a parsed template resource.
@@ -82,7 +82,7 @@ func NewTemplateResource(path string, config Config) ([]*TemplateResource, error
 			return nil, ErrEmptySrc
 		}
 		tr.Src = filepath.Join(config.TemplateDir, tr.Src)
-		templates = append(templates, &tr)
+		templates = append(templates, tr)
 	}
 	return templates, nil
 }
--- a/src/github.com/kelseyhightower/confd/resource/template/resource.go
+++ b/src/github.com/kelseyhightower/confd/resource/template/resource.go
@@ -31,7 +31,7 @@ type Config struct {
 
 // TemplateResourceConfig holds the parsed template resource.
 type TemplateResourceConfig struct {
-	TemplateResource TemplateResource `toml:"template"`
+	TemplateResources []TemplateResource `toml:"template"`
 }
 
 // TemplateResource is the representation of a parsed template resource.
@@ -59,7 +59,7 @@ type TemplateResource struct {
 var ErrEmptySrc = errors.New("empty src template")
 
 // NewTemplateResource creates a TemplateResource.
-func NewTemplateResource(path string, config Config) (*TemplateResource, error) {
+func NewTemplateResource(path string, config Config) ([]*TemplateResource, error) {
 	if config.StoreClient == nil {
 		return nil, errors.New("A valid StoreClient is required.")
 	}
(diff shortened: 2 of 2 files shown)


crandles:
The goal here is to allow a simple install of confd to bootstrap itself based on backend-configured data.

With this, I can create a small RPM to install confd with a simple config and an initial template which will instruct confd to connect to the backend and pull the rest of its configs dynamically. With #314, the templates could also be sourced from the backend, and I would end up with a very dynamic system with one place to make configuration changes.

Without a way to do this, I would need to enforce additional configs and templates via a separate mechanism (Puppet, Ansible, another etcd backed process), and introduce another layer of complexity.

crunchywelch:
I have a similar use case, which could maybe be solved in a similar fashion with a variable in the "dest" value of the config file. We are using uuids to generate service config files for multiple docker containers on the same host.

Starting loosely from the nginx example here https://github.com/kelseyhightower/confd/blob/master/docs/quick-start-guide.md, this:

```
/0491f260-d089-4263-a7f7-74c0144b6874/template
/0491f260-d089-4263-a7f7-74c0144b6874/subdomain
/0491f260-d089-4263-a7f7-74c0144b6874/upstream/app1
/0491f260-d089-4263-a7f7-74c0144b6874/upstream/app2
/b6f0764e-5fb0-4d3c-b559-f76ac2edf8d3/template
/b6f0764e-5fb0-4d3c-b559-f76ac2edf8d3/subdomain
/b6f0764e-5fb0-4d3c-b559-f76ac2edf8d3/upstream/app1
/b6f0764e-5fb0-4d3c-b559-f76ac2edf8d3/upstream/app2
```

would use the value in the "template" key to source /etc/confd/conf.d/{{ template }}.toml, and could use the same variable to define the config file location or name:
```
[template]
prefix = "/nginx"
src = "nginx.tmpl"
dest = "/tmp/{{ template }}.conf"
owner = "nginx"
mode = "0644"
keys = [
  "/subdomain",
  "/upstream",
]
check_cmd = "/usr/sbin/nginx -t -c {{.src}}"
reload_cmd = "/usr/sbin/service nginx reload"
```

This isn't perfect, since it only lets you essentially have one template per UUID, but one could in theory prefix both the UUID and template name with a service name if you wanted (nginx-0491f260-d089-4263-a7f7-74c0144b6874)

The advantage to this is that you are not loading a potentially large number of keys every time with `{{range gets "/nginx/*}}` when, at least in our case, only one service will really be changing at a time, so this should scale a little better.

Thoughts on this?

bsingr:
+1

I do not need all of the proposed capabilites. I am looking for a way to render multiple templates but then call the "reload_cmd" only once.

My goal is to update multiple configuration files for a single application and then restart that application. It's a huge JVM-thing were the restart takes minutes, so I want to call restart only once.

sruon:
+1, any progress?

kelseyhightower:
It seems people want a way to run templates in a specific order. Say run this 10 templates then call reload_cmd at the end. Is that the right way to think about this?

crandles:
I think that's part of it.

The other piece is that those 10 templates are very similar -- either the
confd configs describing them are the same, minus a single identifier, or
the templates are the same as well.

In this sort of a use case it would simplify things if one confd
config/template could be used to write to many destinations.

crandles:
The other piece is that those 10 templates are very similar -- either the
confd configs describing them are the same, minus a single identifier, or
the templates are the same as well.

In this sort of a use case it would simplify things if one confd
config/template could be used to write to many destinations.

kelseyhightower:
@crandles What about providing a way to provide a list of src/dest mappings? Something like this:

```
[template]

[[targets]]
src = "nginx.conf.tmpl"
dest = "/etc/nginx/nginx.conf"

[[targets]]
src = "www.example.com.conf.tmpl"
dest = "/etc/nginx/conf.d/www.example.com.conf"

keys = [
    "/nginx/",
]

check_cmd = "/usr/sbin/nginx -t -c {{.src}}"
reload_cmd = "/usr/sbin/service nginx reload"
```

crandles:
I think that might be helpful. 

In that case would the "keys" be utilized across all targets?

kelseyhightower:
Now that I had more time to think about this I don't like where this is headed. confd is a simple tool designed to watch some keys in a config database and generate a single template. I know it's limiting, but it solves the 80% use case.

While I like this idea, I think it's the wrong direction for confd. I'm closing this PR out of the sake of simplicity.

crandles:
Understood. I think I've found somewhere else to manage the complexity -- I appreciate the consideration.

PR #323, opened 2015-08-18
Add IP lookup template function

cloudnautique:
This PR adds a template function that aliases net.LookupIP. This
is helpful in situations where you need to know an IP with the environment
backend and do not know IP info ahead of time.


cloudnautique:
I think #333 covers my use case and the sort capability makes sense. I'll close this for now.

PR #328, opened 2015-08-22
On start, allow etcd backend to try reconnect if error

bnadim:


src/github.com/kelseyhightower/confd/backends/etcd/client.go
```diff
@@ -28,10 +30,23 @@ func NewEtcdClient(machines []string, cert, key string, caCert string) (*Client,
 	}
 	// Configure the DialTimeout, since 1 second is often too short
 	c.SetDialTimeout(time.Duration(3) * time.Second)
-	success := c.SetCluster(machines)
-	if !success {
-		return &Client{c}, errors.New("cannot connect to etcd cluster: " + strings.Join(machines, ","))
+
+	maxConnectAttempts := 10
```
bacongobbler: this should stay as 1 for now so others aren't caught by this.

bnadim: 1 is really low if you have to wait for an etc service to come up.
This should only happen on launch of confd.
Maybe we should crash the process if it doesn't connect because anyway confd won't do anything if it doesn't connect the first time ?

bacongobbler: Yes. What I mean is that the behaviour should stay the same for now, but you should be able to tune (with a flag) how many attempts can be made to connect to etcd before it errors out.

We should probably make this configurable across all backends as well, as this would be a handy feature for more than one backend.

bnadim: Ok i will try to update it as soon as i have some time.

kelseyhightower:
I have not heard any more issues related to this. Closing.

PR #329, opened 2015-08-25
Allow proper use of etcd proxies.

### LINKED ISSUE #319: confd bypasses local proxy, all instances connect directly to single cluster member
I'm doing some scale testing of my application with a 5-node etcd cluster and 500+ client hosts.  I have an etcd proxy on each of my client hosts.  confd seems to initially connect to the local proxy but then, on subsequent requests, it connects directly to a node in the cluster.  Worse, all the confd instances seem to pick the same cluster node to talk to, which causes excessive load on that node and makes it unresponsive.

```
# /confd --version
confd 0.9.0
```

Lukasa:
Resolves #319.

This pull request adds a `--no-discover` flag that applies only when using the etcd backend. This flag prevents confd from discovering the whole cluster, and so ensures that it only uses the proxy.

This change may be incomplete: right now with the change in this form you need to provide the etcd authority in the form `http://host:port` when using the `--no-discover` flag. If that's problematic, we can copy go-etcd's `Client.createHttpPath` method, which can add a scheme to the front of the authority if needed. Thoughts?

### DIFF
--- a/src/github.com/kelseyhightower/confd/backends/client.go
+++ b/src/github.com/kelseyhightower/confd/backends/client.go
@@ -36,7 +36,7 @@ func New(config Config) (StoreClient, error) {
 	case "etcd":
 		// Create the etcd client upfront and use it for the life of the process.
 		// The etcdClient is an http.Client and designed to be reused.
-		return etcd.NewEtcdClient(backendNodes, config.ClientCert, config.ClientKey, config.ClientCaKeys, config.NoDiscover)
+		return etcd.NewEtcdClient(backendNodes, config.ClientCert, config.ClientKey, config.ClientCaKeys, config.NoSync)
 	case "zookeeper":
 		return zookeeper.NewZookeeperClient(backendNodes)
 	case "redis":
--- a/src/github.com/kelseyhightower/confd/backends/client.go
+++ b/src/github.com/kelseyhightower/confd/backends/client.go
@@ -36,7 +36,7 @@ func New(config Config) (StoreClient, error) {
 	case "etcd":
 		// Create the etcd client upfront and use it for the life of the process.
 		// The etcdClient is an http.Client and designed to be reused.
-		return etcd.NewEtcdClient(backendNodes, config.ClientCert, config.ClientKey, config.ClientCaKeys)
+		return etcd.NewEtcdClient(backendNodes, config.ClientCert, config.ClientKey, config.ClientCaKeys, config.NoDiscover)
 	case "zookeeper":
 		return zookeeper.NewZookeeperClient(backendNodes)
 	case "redis":
--- a/src/github.com/kelseyhightower/confd/backends/config.go
+++ b/src/github.com/kelseyhightower/confd/backends/config.go
@@ -7,6 +7,7 @@ type Config struct {
 	ClientCert   string
 	ClientKey    string
 	BackendNodes []string
+	NoDiscover   bool
 	Scheme       string
 	Table        string
 }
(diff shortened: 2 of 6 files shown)


Lukasa:
I've actually just gone ahead and added some code that automatically prepends the scheme.

bacongobbler:
I think with etcdctl they call this `--no-sync` so it might make sense to rename it as such.

Lukasa:
Seems reasonable to me.

markhowells:
Fixes #354 for me.

geku:
+1

kelseyhightower:
This should be fixed with the move to the new etcd client libs. Please feel free to reopen if that is not the case.

jeromegn:
This is not resolved. I'm still hitting that issue. Any URL options I need to pass? I don't see any relevant CLI flags. Using 0.11

HeavyHorst:
The new etcd client lib is included in the 0.12 pre-releases. 0.11 still
uses the old Library. Can you please test this with one of the 0.12
releases ?

Jerome Gravel-Niquet <notifications@github.com> schrieb am Di., 27. Sep.
2016, 22:57:

jeromegn:
@HeavyHorst oh, you're right. I just tried it and it seems to work, mostly.

The initial lookup works, as well as the subsequent ones when watching. Which is what troubles me given my logs are being spammed with:

```
2016-09-27T18:09:29-04:00 jgn.local confd[78591]: ERROR client: etcd cluster is unavailable or misconfigured
2016-09-27T18:09:37-04:00 jgn.local confd[78591]: ERROR client: etcd cluster is unavailable or misconfigured
2016-09-27T18:09:45-04:00 jgn.local confd[78591]: ERROR client: etcd cluster is unavailable or misconfigured
2016-09-27T18:09:53-04:00 jgn.local confd[78591]: ERROR client: etcd cluster is unavailable or misconfigured
2016-09-27T18:10:01-04:00 jgn.local confd[78591]: ERROR client: etcd cluster is unavailable or misconfigured
2016-09-27T18:10:09-04:00 jgn.local confd[78591]: ERROR client: etcd cluster is unavailable or misconfigured
2016-09-27T18:10:17-04:00 jgn.local confd[78591]: ERROR client: etcd cluster is unavailable or misconfigured
```

It works though, which is conflicting with those messages.

It's worth noting I current use a 2 proxy setup, through https, I provide a client-ca-keys argument, a username and a password.

HeavyHorst:
I noticed that the Error occurs every 8 seconds. The etcd-watcher uses http
long polling to track changes. Is it possible that one of your proxies had
a connection timeout of 8 seconds configured and closes the connection
because of that ?

Jerome Gravel-Niquet <notifications@github.com> schrieb am Mi., 28. Sep.
2016, 00:23:

jeromegn:
I believe I'm hitting this issue: https://github.com/coreos/etcd/issues/3405

I've looked at the proxies' logs and all I see are those premature closed connections (at the same interval you mentioned):

```
2016-09-28 11:38:38.350727 I | proxy: client 127.0.0.1:51677 closed request prematurely
2016-09-28 11:38:46.351517 I | proxy: client 127.0.0.1:52204 closed request prematurely
2016-09-28 11:38:54.352242 I | proxy: client 127.0.0.1:52714 closed request prematurely
2016-09-28 11:39:02.353108 I | proxy: client 127.0.0.1:53205 closed request prematurely
```

Could be an issue with compose.io though (that's where my etcd is.)

jeromegn:
Looks like this may be related too: https://github.com/coreos/etcd/issues/3894 (which is fixed by: https://github.com/coreos/etcd/pull/4254)

:thinkingface:

HeavyHorst:
@jeromegn i can reproduce the error with a newly created compose.io etcd cluster.
i also tested this with github.com/HeavyHorst/remco, its working fine there.

I think it should be sufficient to bump the etcd client to the latest version.

PR #330, opened 2015-08-27
Improve efficiency of etcd watches.

### LINKED ISSUE #326: Tighten scope for templates in watch mode.
Currently when confd is running in watch mode it will watch an entire key prefix for changes. Each time that watch pops, confd will re-query each of the keys and then render out the template.

If the prefix happens to be quite high up the tree compared to the keys, and the keyspace is heavily populated, this can lead to confd doing a lot of work it does not need to do, because confd does not check whether or not any of the keys it's interested in has changed, even when it could otherwise do so.

There are two possible changes we could make here to improve this situation.

The simplest one is to extend the `StoreClient` interface by adding an extra argument to `WatchPrefix` that provides the specific keys confd cares about to the `StoreClient` implementation. Implementations that choose to can then optionally filter whether they return from a watch based on whether those keys have changed. I'd be open to providing a pull request that implements this change in the `watchProcessor` and the etcd backend.

This approach will reduce the amount of work confd does, which is nice, but still leads to more load being put on etcd than really needs to be because a large number of watches may pop that don't need to.

The more complicated one is to set up watches on each key directly. This is more complicated only because it involves a bit of extra code in some weird places, but is otherwise probably fairly simple. Whether this reduces load on etcd compared to the first option depends a bit on the use-case: lots of separate watches ties up file descriptors and other resources, so depending on the use-case this might not be a clear win.

I'd also be happy to whip-up an implementation of this change, again focusing on etcd.

Do you have a preference as to which approach would fit confd better? At the moment I'm leaning towards the first, but I could go either way.

Lukasa:
Resolves #326.

This change is a proposal to improve the performance problems I highlighted in #326. See that issue for more discussion.

### DIFF
--- a/src/github.com/kelseyhightower/confd/backends/client.go
+++ b/src/github.com/kelseyhightower/confd/backends/client.go
@@ -18,7 +18,7 @@ import (
 // key/value pairs from a backend store.
 type StoreClient interface {
 	GetValues(keys []string) (map[string]string, error)
-	WatchPrefix(prefix string, waitIndex uint64, stopChan chan bool) (uint64, error)
+	WatchPrefix(prefix string, waitIndex uint64, stopChan chan bool, keys []string) (uint64, error)
 }
 
 // New is used to create a storage client based on our configuration.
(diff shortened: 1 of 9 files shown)


kelseyhightower:
@Lukasa Thanks for the patch, but this will require a little more testing on my part.

sruon:
+1

kelseyhightower:
This has been fixed on the master branch. Please test.

PR #331, opened 2015-08-28
Add autoscaling backend for EC2 Auto Scaling Groups

riccardofreixo:
Added a backend for EC2 Auto Scaling Groups with the latest version of the AWS SDK for Go.

It picks up all these from the instances of a named Auto Scaling Group:
 * Private IP address
 * Private DNS name
 * Public IP address
 * Public DNS name

And exposes them under the keys /privateIps/N, /privateDnsNames/N, /publicIps/N, /publicDnsNames/N, where N is an index starting at 0.

It gets AWS credentials from all sources the SDK gets them from. It can be used as below:
confd -onetime -backend autoscaling -asg "My Autoscale Group"  -awsregion eu-west-1

The region can also be read from the AWS_REGION env var.

Since I've added the latest version of the SDK, if needed I can migrate the dynamo backend to use the latest version of the SDK too.


kelseyhightower:
This change is out of scope for confd. The hard and fast rule here is that I only want to support key/value data stores like etcd and consul. For your use case you will need to push this kind of data to one of the supported backends and query the information from there.

PR #333, opened 2015-09-09
IP dns resolving function for templates

jnummelin:
PR adds IP lookup function for templates. SImilar to #323 with small but significant difference. This version sorts the IP list so that unnecessary config reloads can be avoided. 

As we are using this on Docker and using Alpine as our base image there's also builder image for Alpine package building.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+bin
+vendor
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,3 @@
 bin
 vendor
+pkg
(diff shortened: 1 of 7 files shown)


bketelsen:
nice

cloudnautique:
Cool! I'm looking forward to the iplookup function.

One suggestion with the Alpine builder would be to create a static go binary that can be used in multiple base images. While Alpine is great, and one of my favorites, its also pretty common to use busybox and scratch. Might also work to use the golang image and cross compile all of the binaries so that its not just a Linux specific approach.

kelseyhightower:
@jnummelin  break out the alpine build image to a different commit please, then I can merge.

Issue #73, opened 2014-05-23
Support for environment variables when processing templates


mentalblock:
It would be convenient if confd supported interpolation of templates by using values from environment variables. The following use cases would take advantage of this feature.
- Local development on a workstation which does not have a local etcd or consul available can make use of the confd templates. When a configuration service is available, one can switch to a confd configuration that will make use of it.
- Less time needed for setup and configuration if we just want testing of configuration templates. Iterating configuration templates will only require setting environment variables.

bilalaslam:
:+1: on docs

bacongobbler:
There's simple documentation at https://github.com/kelseyhightower/confd/blob/master/docs/quick-start-guide.md#environment-variables

linbynd:
In the documentation the example is using the "getv". For pulling Environment variables. use getenv.

Ex:
```sh
[template]
src = "myconfig.conf.tmpl"
dest = "/tmp/myconfig.conf"
keys = [
    "USER",
    "HOME",
]

[myconfig]
USER = {{getenv "USER"}}
HOME = {{getenv "HOME"}}
```
info in https://github.com/kelseyhightower/confd/blob/master/docs/templates.md

duffqiu:
Can I use environment together with etcd something like fleet? because I wan to set etcd keys dynamically.

for example

[template]
src = "haproxy.cfg.tmpl"
dest = "/etc/haproxy/haproxy.cfg"
keys = [
        "/${haproxy1}/apps/www/app1"
]
reload_cmd = "echo haproxy soft restarting"

bacongobbler:
AFAIK you can't use environment variables in the toml files.

duffqiu:
Can add this as a new feature?

bacongobbler:
Sure :)

duffqiu:
I am not a developer for go. But I try another workaround solution for that. First, I write a bash to support environment to generate the template in the folder of confd. and the confd can generate the configuration file for the applications such as haproxy. you can have a look at [confd-init](https://github.com/duffqiu/confd-init) if you are interesting in it.

knepperjm:
Just found this.  <3 u Kelsey Hightower.

PR #344, opened 2015-10-12
add support for dynamic keys, using environment variables

davidquarles:
This is quick hack, based on the behavior described in #73 and #310.
Example / our use case:
```toml
[template]
src = "database.php.tmpl"
dest = "/var/www/site/application/config/database.php"
keys = [
    "/cluster/${CLUSTER_ID}/mysql/master/internal",
    "/cluster/${CLUSTER_ID}/mysql/slave/internal",
]
```
Template:
```php
<?php  if (!defined('BASEPATH')) exit('No direct script access allowed');

{{ $clusterID := getenv "CLUSTER_ID" }}

$db['default']['hostname'] = "{{ getv (printf "/cluster/%s/mysql/master/internal" $clusterID) }}"
// ...

$db['slave'] = $db['default'];
$db['slave']['hostname'] = "{{ getv (printf "/cluster/%s/mysql/slave/internal" $clusterID) }}"
?>
```

### DIFF
--- a/src/github.com/kelseyhightower/confd/resource/template/resource.go
+++ b/src/github.com/kelseyhightower/confd/resource/template/resource.go
@@ -70,6 +70,7 @@ func NewTemplateResource(path string, config Config) (*TemplateResource, error) 
 		return nil, fmt.Errorf("Cannot process template resource %s - %s", path, err.Error())
 	}
 	tr := tc.TemplateResource
+	tr.expandEnv()
 	tr.keepStageFile = config.KeepStageFile
 	tr.noop = config.Noop
 	tr.storeClient = config.StoreClient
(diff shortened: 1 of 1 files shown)


davidquarles:
I can totally amend the docs, if this is something you're interested in merging.  We could also explicitly specify which environment vars we're looking for in the config and limit expansion to those, but I figured this was a step in the right direction.  Let me know what you think.  Thanks!

kelseyhightower:
While this seems to work for some use cases it increases the complexity of confd. The original goal of confd was to be simple to understand and reason about after deployment. This change will introduce dynamic keys into the mix and goes against the spirt of confd a bit, where I want to keep things as explicit as possible.

I don't see a limit to this feature, and I will soon see requests to pull vars from all the backends so they can be used to bind the dynamic keys. Please understand my desire to keep confd super simple and serve as a simple bridge between applications and configuration stores.

PR #347, opened 2015-10-23
add test for the path has same prefix bug caused by memkv.

wuranbo:
the bug is solved by https://github.com/kelseyhightower/memkv/pull/7
so this PR should be merged later when the issue is fixed.


wuranbo:
close caused shoud be merged after the issue is fixed.

PR #348, opened 2015-10-27
Add lookupSRV function.

jburnham:
Adds support for fetching SRV records and using them in a template. Would go great in conjunction with https://github.com/kelseyhightower/confd/pull/333

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -136,6 +136,19 @@ key: {{toLower "Value"}}
 
 Returns an map[string]interface{} of the json value.
 
+### lookupSRV
+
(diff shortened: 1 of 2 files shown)


kelseyhightower:
@jburnham Now that #333 has been merged, this will need a rebase. I think we really need to figure out a better way of adding template functions.

jburnham:
Rebased, It looks like the failing test is because of the lookupIP function. TravisCI seems to not have ipv6 so resolving localhost is only 127.0.0.1.

kelseyhightower:
@jburnham Seems there is still a conflict. I fixed a bunch of test so things should be all green now.

jburnham:
Rebased again.

kelseyhightower:
@jburnham Thanks!

PR #363, opened 2015-11-30
Feature new template func add

### LINKED ISSUE #343: add and subtract template functions
It would be nice to be able to increase or decrease numbers in templates.
My use case is that when a system is a cache layer and registers itself in zookeeper, it should be able to access the upper/lower layers by configuration. Calling {{ add $layerId "1" }} would simply do the trick.

I have a PR ready if you're up for it.

### LINKED ISSUE #576: Add template functions for the basic arithmetic operations
I'd like to be able to do some lightweight calculations within confd templates. This would be useful for my confd template for elasticsearch. The elasticsearch guys recommend to configure the option for the minimum amount of masters based on the rule: `(master_eligible_nodes / 2) + 1`
As I need a list of all master hosts anyway within the template it would be nice if I could use the length of this to calculate the value for the minimum amount of masters instead of passing an additional value to the template.
Especially for this case it's necessary to have a function for the integer division.

aquam8:
No impact on master. No backward compatibility issues.
Just added a few arithmetic operator to the templating engine. I have found the modulus operator very important for confd.
Please consider merging this pull request.
Cheers!

This fixes #343, fixes #576

### DIFF
--- a/src/github.com/kelseyhightower/confd/resource/template/template_funcs.go
+++ b/src/github.com/kelseyhightower/confd/resource/template/template_funcs.go
@@ -23,6 +23,11 @@ func newFuncMap() map[string]interface{} {
 	m["contains"] = strings.Contains
 	m["replace"] = strings.Replace
 	m["add"] = func(a, b int) int { return a + b }
+	m["sub"] = func(a, b int) int { return a - b }
+	m["div"] = func(a, b int) int { return a / b }
+	m["mod"] = func(a, b int) int { return a % b }
+	m["mul"] = func(a, b int) int { return a * b }
+	m["modBool"] = func(a, b int) bool { return a%b == 0 }
 	return m
 }
 
--- a/src/github.com/kelseyhightower/confd/resource/template/template_funcs.go
+++ b/src/github.com/kelseyhightower/confd/resource/template/template_funcs.go
@@ -22,6 +22,7 @@ func newFuncMap() map[string]interface{} {
 	m["toLower"] = strings.ToLower
 	m["contains"] = strings.Contains
 	m["replace"] = strings.Replace
+	m["add"] = func(a, b int) int { return a + b }
 	return m
 }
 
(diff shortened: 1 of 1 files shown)


robinsq:
Any chance we can merge this. I agree that arithmetic operations are useful.

kelseyhightower:
This is a hard one to consider for merging because I wonder if we are heading down the right path with this. Do we want to expose a full scripting environment to confd configs? Are we getting away from simplicity?

kelseyhightower:
@aquam8 Can you rebase this? I'll merge it once you do.

corebug:
@kelseyhightower basically limited number of arithmetic functions doesn't look like a lot of complexity, but it really extends confd functionality, especially when it comes to huge complex templates that require additional logic.

sporokh:
@corebug @kelseyhightower Math operations are really useful in some cases to generate it via confd.

VitaliiTurovets:
@sporokh @kelseyhightower rebased version of this PR is here: https://github.com/kelseyhightower/confd/pull/515.

okushchenko:
Fixes #343 #576

JPvRiel:
> This is a hard one to consider for merging because I wonder if we are heading down the right path with this. Do we want to expose a full scripting environment to confd configs? Are we getting away from simplicity?

A challenge for me is coming from other templating engines like jinja and finding golang template processing highly restrictive. Golang text templating assumes that use case is someone who is running a golang program. Now we're a little stuck here in a use case where we cannot run and set variables or data structures via go and try to force/include extra logic into the templating...

I noticed the consul project was also hosting a whole bunch of functions to go around the limitations.

okushchenko:
@JPvRiel this PR is already merged, do you need any other functions in confd?

JPvRiel:
> @JPvRiel this PR is already merged, do you need any other functions in confd?

Sorry, not requesting an extra functions. Was just a latent comment about the dilemma of trying to bundle too many features/programming logic in a templating language itself. And guessing that golang templating wasn't created with external use cases in mind, hence it being restrictive/avoiding duplicating programming language features like arithmetic functions/operators.

idonin:
Missing `docs/templates.md` updates for this PR.

Issue #146, opened 2014-10-07
ACL support for consul backend


provonet:
Hi,

Consul 0.4.0 supports acl's, are there any plans to support this in the consul backend ? 

cheers,

Johan

m1no:
This feature would help a lot to secure consul setups with confd +1

juicedM3:
+1

juicedM3:
I was hoping that the auth-token config param would at least pass the master token to consul but it doesn't. Just master token support would be a good start.

juicedM3:
I've dug around the code and found if I set the ENV variable CONSUL_HTTP_TOKEN, I can pass the master token.

juicedM3:
I've forked the code and have a working example (https://github.com/juicedM3/confd). I need to write tests and create a pull request.

juicedM3:
Pull request https://github.com/kelseyhightower/confd/pull/369.

rares-aioanei:
Is there any update regarding the status of this feature?

bacongobbler:
Feel free to take a crack at a PR.

billyteves:
still no update i guess?

bacongobbler:
Project dev has essentially stalled for the last two years, so yes.

miaoyc1989:
It is now 2020, where is features

PR #369, opened 2015-12-14
 ACL support for consul backend #146 

juicedM3:
This pull request allows you to pass in individual ACL tokens for keys. For example:

```
[template]
src = "config.yaml.tmpl"
dest = "/opt/bla/config.yaml"
prefix = "/bla"
token = '00112233AABB'
keys = [
"key1",
"key2",
]
```


kelseyhightower:
I think a better way to handle this is to consider using the go context library. That way we can allow each backend to carry metadata specific to that backend. Thoughts?

PR #374, opened 2016-01-05
Add more debugging output for onetime and env.

relistan:
When setting up a new installation running in `onetime` mode and using the `env` backend I ran into a number of issues that produced no error or debugging output. I ended up modifying the code to tell me what's happening. This PR add some more logging in areas where you are currently blind and hopefully will benefit anyone else facing a similar issue.


bacongobbler:
LGTM

PR #380, opened 2016-01-25
redis: try to recover from closed connections

### LINKED ISSUE #373: Redis backend cannot recover from node failure
I'm using `confd` with the `redis` backend and wanted to see if `confd` could get back to work after the one (and only) `redis` node disappears for a couple of minutes. 

Unfortunately, when Redis was back up, the `confd` `redis` backend was apparently still trying to use the original network connection (now closed). I got the following log message over and over again:
```
/opt/bin/confd: ERROR write tcp 172.17.0.1:56142->172.17.0.1:6379: use of closed network connection
```

I had hoped, that confd/the redis backend would try to establish a new connection when its current one fails.

SealedSun:
Make the redis client test its connection before each GetValues operation with a
`PING` command. If the connection is dead, try to connect to one of the configured redis nodes. Only performs a single reconnect attempt. For my use case, I can just wait until confd tries to apply configuration templates again.

Fixes #373 as far as I'm concerned.

Weaknesses:
 * _nodes are always contacted in the same order_, should probably randomize/hash order or something? 
 * _one re-connection attempt for each template, not just once for each time the interval fires_. We'd probably need a more sophisticated backend/client interface to improve on that.

I don't have much experience with go. I'm happy to improve/correct the PR if necessary.

### DIFF
--- a/src/github.com/kelseyhightower/confd/backends/redis/client.go
+++ b/src/github.com/kelseyhightower/confd/backends/redis/client.go
@@ -6,16 +6,19 @@ import (
 	"os"
 	"strings"
 	"time"
+	"github.com/kelseyhightower/confd/log"
 )
 
 // Client is a wrapper around the redis client
 type Client struct {
-	client redis.Conn
+	client   redis.Conn
+	machines []string
 }
 
-// NewRedisClient returns an *redis.Client with a connection to named machines.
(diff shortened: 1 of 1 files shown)


kelseyhightower:
I can refactor this before the next release. Thanks for the contribution.

PR #383, opened 2016-01-28
Added support for Vault backend

odedlaz:
added support for vault backend

current implementation supports the following auth backends: token, github, app-id, userpass
furthermore, because vault stores keys in json format, some tweaks had to be done.
if the json has only one key, it's value is used. for example:
``` bash
$ vault write secret/foo some_value=bar
```
will be treated as a key-value of: ```secret/foo: "bar"```

if there are multiple keys, than the json encoded key will be saved at the root
and will be flattened in order to support ```gets``` and ```getvs``` and ```json``` keywords. for example:
``` bash
$ vault write secret/foo some_value=bar some_other_value=hello
```
will produce multiple keys:

* ```secret/foo```
```json
{
   "some_value": "bar",
   "some_other_value": "hello"
}
```
* ```secret/foo/some_value``` with the value: ```"bar"```
* ```secret/foo/some_other_value``` with the value: ```"hello"```


sylus:
This commit is so very exciting, thanks for the awesome work!

odedlaz:
no problems :smile:

PR #395, opened 2016-02-22
updated Dockerfile.build.alpine

HeavyHorst:
the old Dockerfile was no longer working because of the switch to the go 1.6 vendoring support in favor of gb.

### DIFF
--- a/Dockerfile.build.alpine
+++ b/Dockerfile.build.alpine
@@ -1,5 +1,4 @@
 FROM alpine:3.3
-MAINTAINER Rene Kaufmann <kaufmann.r@gmail.com>
 
 ENV GOPATH /go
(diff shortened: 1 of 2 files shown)


Dockerfile.build.alpine
```diff
@@ -1,9 +1,13 @@
-FROM gliderlabs/alpine:3.2
-MAINTAINER Jussi Nummelin <jussi.nummelin@gmail.com>
-RUN apk-install bash go bzr git mercurial subversion openssh-client ca-certificates
+FROM alpine:3.3
+MAINTAINER Rene Kaufmann <kaufmann.r@gmail.com>
```
bacongobbler: We should just remove this MAINTAINER tag or everyone's just going to be updating it every time a patch is made.

PR #398, opened 2016-02-24
Add support for redis password authentication

### LINKED ISSUE #317: Redis backend AUTH support
Hello, 
As far as I can see in https://github.com/kelseyhightower/confd/blob/master/src/github.com/kelseyhightower/confd/backends/client.go#L42L43 there's no AUTH support for the REDIS backend (ref. http://redis.io/commands/AUTH)

This can be very useful with hosted Redis instances (Amazon ElastiCache, Azure Redis Cache etc...) or when is not possible to apply firewall rules.

Is there already a plan to implement it? 'redigo' package is already compatible but my Go is still too newbie for a pull request, I'm sorry :)

ericbarch:
This pull request updates redigo to the latest and incorporates optional use of the client key config parameter to use for redis password auth. Thanks goes to israelchen for initially tackling this.

This pull request will bring the code up to date with confd's master.

This should resolve #317.


kelseyhightower:
@ericbarch Awesome!

PR #403, opened 2016-02-27
improved etcd watch support

HeavyHorst:
I think it is not necessary to run 

```
ctx, cancel := context.WithCancel(context.Background())
cancelRoutine := make(chan bool)
defer close(cancelRoutine)

go func() {
    select {
    case <-stopChan:
        cancel()
    case <-cancelRoutine:
        return
    }
}()
```

all the time in the for loop. 
This will waste resources in the case that keys with a key prefix we don't care about are changed in a high frequency.

I made a few simple changes to the watchPrefix method to improve these behavior.

---

This change is [Reviewable](https://reviewable.io/reviews/kelseyhightower/confd/403)

### DIFF
--- a/backends/etcd/client.go
+++ b/backends/etcd/client.go
@@ -115,49 +115,65 @@ func nodeWalk(node *client.Node, vars map[string]string) error {
 	return nil
 }
 
+type watchResponse struct {
+	waitIndex uint64
+	err       error
+}
+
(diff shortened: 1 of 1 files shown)


bacongobbler:
I've gone ahead and made a v0.12.1 release on my fork that includes this change. See [the v0.12.0](https://github.com/bacongobbler/confd/releases/tag/v0.12.0) and [v0.12.1 release notes](https://github.com/bacongobbler/confd/releases/tag/v0.12.1).

PR #410, opened 2016-03-06
extended getenv & map template functions

odedlaz:
### getenv
extended the `getenv` template function to allow default values.
useful when you want to have a default value if nothing was supplied, for instance:
`{{ getenv "HOST_IP" "127.0.0.1" }}`

### map
added a new `map` function that allows to create maps inside a template. for instance:
`{{$endpoint := map "name" $name "private_port" $private_port "public_port" $public_port}}`

this is very useful if you have a defined sub-templates and you want to pass multiple values to them.
the idea & code was written by **tux21b** at [stackoverflow](https://stackoverflow.com/questions/18276173/calling-a-template-with-several-pipeline-parameters/18276968#18276968)

### DIFF
--- a/resource/template/template_funcs.go
+++ b/resource/template/template_funcs.go
@@ -20,7 +20,7 @@ func newFuncMap() map[string]interface{} {
 	m["jsonArray"] = UnmarshalJsonArray
 	m["dir"] = path.Dir
 	m["map"] = CreateMap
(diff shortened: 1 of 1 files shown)


kelseyhightower:
@odedlaz LGTM.

micahhausler:
It'd be great to see this in a release!

Issue #414, opened 2016-03-09
Multiple Backend Sources for same template?


fbrbovic:
Is it possible to have access to multiple backend sources for the same template?

For example  we have both Rancher metadata and vault, so in our configuration file we need access to both to be able to build the config file.

bacongobbler:
there's no way to enable two backends at once, unfortunately.

fbrbovic:
Having such capability is something that's definitely needed, especially when dealing with passwords, which you just don't want to be passing around unencrypted inside normal discovery service tools or as environment vars.

michaellopez:
We have the same use case, Rancher + env or Rancher + vault. So +1 here :)

bacongobbler:
@michaellopez would `{{ getenv "MY_ENVVAR" }}` work for you in tandem with the rancher backend? 

Having to support multiple backends on a single template is not something we'd likely implement as it would require an entire project re-architecture and quite a few backwards-incompatible changes for a single feature. confd currently works on a single backend for all templates, configured by a runtime option flag.

michaellopez:
@bacongobbler Thanks for the info, I'll try it out and get back to you. I don't see how I missed that, I've read the docs a million times :)

Is there a possibility to run two instances of confd or two passes or similar to get around the need for multiple backends? Is there a recommended way to solve this?

bacongobbler:
Running two instances of confd would just make one template overwrite the other. The recommended way to solve this would be to import all the keys from one backend to the other and just use that as the backend.

fbrbovic:
Importing keys from one backend to another would not work in case of a Vault backend, since we are dealing with encrypted data, it would defeat the whole purpose if we had to expose the secrets.

fbrbovic:
One option is to switch to Consul from Rancher meta data service and use Consul-Templates https://github.com/hashicorp/consul-template. It works similar to confd but reads from both consul and vault.   The only problem is now we need some kind of sync tool to sync up Rancher metadata to consul. 

There is also an open ticket in rancher for it's own secret management. Not sure if its moving anywhere.

bacongobbler:
Yes, if you're consuming the entire hashicorp stack then it makes sense to use consul-template because it has intimate knowledge of those two backends. confd is backend-agnostic so it only works against a single backend at a time and uses it as a key/value store rather than a secret management tool.

fbrbovic:
I really needed this feature for my own purposes so I ended up extending confd to support a 2nd backend of choice

[...]

PR #417, opened 2016-03-21
Enabling 2nd Backend git 

fbrbovic:
This pull request allows for having confd poll two backends at the same time and provide the data from both in the same templates.

To use it : 
set `-backend2=` to backend of your choice
example:
```
confd -backend=etcd -backend2=vault -auth-type2=token -auth-token2=${AUTH_TOKE} -node2=http://10.22.1.102:8200
```
set any additional flags related to the backend2  (flags which end with '2' example `-node2=` )

set backend2 keys in the toml file, by setting `keys2=`
example:
```
[template]
src = "test.tmpl"
dest = "/tmp/test.conf"
mode = "777"
keys = [
  "/message",
]
keys2 = [
  "/secret/api/external/docker_hub",
]
```

use the backend 2 vars in the template file by prefixing the keys with `backend2:`
example: 
```
{{getv "backend2:/secret/api/external/docker_hub/username"}}
```

This request was stemmed from issue : https://github.com/kelseyhightower/confd/issues/414

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 bin/
+.idea
--- a/backends/etcd/client.go
+++ b/backends/etcd/client.go
@@ -138,7 +138,6 @@ func (c *Client) WatchPrefix(prefix string, keys []string, waitIndex uint64, sto
 				return
 			}
 		}()
-
 		resp, err := watcher.Next(ctx)
 		if err != nil {
 			switch e := err.(type) {
--- a/confd.go
+++ b/confd.go
@@ -30,6 +30,19 @@ func main() {
 	}
 
 	templateConfig.StoreClient = storeClient
+
+	// 2nd backend setup
+	if backendsConfig2.Backend != "" {
+		log.Debug("Preparing 2nd Backend Config ")
+		storeClient2, err2 := backends.New(backendsConfig2)
+
+		if err2 != nil {
+			log.Fatal(err2.Error())
+		} else {
+			templateConfig.StoreClient2 = storeClient2
+		}
+	}
+
 	if onetime {
 		if err := template.Process(templateConfig); err != nil {
 			log.Fatal(err.Error())
--- a/config.go
+++ b/config.go
@@ -44,11 +44,27 @@ var (
 	table             string
 	templateConfig    template.Config
 	backendsConfig    backends.Config
+	backendsConfig2   backends.Config
 	username          string
 	password          string
 	watch             bool
 	appID             string
 	userID            string
+	// 2nd backend config
+	authToken2         string
+	authType2          string
+	backend2           string
+	basicAuth2         bool
+	clientCaKeys2      string
+	clientCert2        string
+	clientKey2         string
+	nodes2             Nodes
+	prefix2            string
+	table2             string
+	username2          string
+	password2          string
+	appID2             string
+	userID2            string
 )
 
 // A Config structure is used to configure confd.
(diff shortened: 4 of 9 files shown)


michaellopez:
I think this looks promising

odedlaz:
The idea is awesome, but I think we need to find a generic way of doing this instead of allowing just one more backend. plus, there's a lot of duplication in the code which will make changing and bugs fixes difficult in the future.

fbrbovic:
@odedlaz Yes I agree there is definitely a lot that can be done to generalize adding of additional backends. I needed a simple solution to enable a 2nd backend right away for my use case.
I've considered working on a more generalize way but it would require a lot more work since we would also need to ensure that it would also be backwards compatible with existing flags. 

One option I considered was to keep the same flags. Example: ```--backend=```
but enable some kind of namespacing/aliasing capability inside the flag values like : ```--backend="backend1:etcd,backend2:vault,backend3:rancher"```, but it would have taken me a lot longer to implement, and I didn't have time. 

Question is also beyond 2nd or possible third backend, how many uses cases would need more then 2 or 3 simultaneous backends to justify added complexity?

bacongobbler:
A couple of problems I see with this approach are key collision, maintainability and the added complexity of this PR.

For key collision, keys are now namespaced by their backend (`getv "backend1:/foo"`) in this PR. Not only is this backwards-incompatible, but this also means that the templates must have intimate knowledge of the way confd was configured to fetch their keys. The key paths are meant to be a filesystem-like key/value store similar to etcd's key/value layout. Even if we reverted this to make all keys added to the root path (`/`), we would then have the issue of collision detection in two backends that have the same path (e.g. etcd has `/foo` and vault has `/foo`). This alone is a huge issue that would need to be properly resolved.

To address maintainability, if we allow N backends to be configured, what's wrong with configuring N+1? If we're already going down this approach, a user should be able to configure any number of backends. 

From #414 it just looks like the problem's more about how to retrieve values from vault securely while also storing non-critical information in another backend, not retrieving values from more than one backend at a time. I don't see how it'd be useful to read keys from both etcd and stackengine, for example.

To summarize I don't think this is the right way to solve #414. It adds a significant amount of added complexity to solve one use case, which doesn't seem like it's worth the hassle to maintain IMO.

fbrbovic:
@bacongobbler  As far as I understand templates always need to have intimate knowledge of how confd is configured because confd allows to set a prefix and that changes the way vars are accessed. If you change the prefix then templates will now not work just like if you were to swap the backends, as you described it. 

Either way backend swapping problem can be solved if we were to implement using the name of the backend as the namespace for the variable instead of the word "backend". 
So we would have ```getv "vault:/foo"```  instead of ```getv "backend1:/foo"```.   This way order the backends doesn't matter. Of course we would not be able to use two same backend types together. 

Also I don't see how using a namespace  (name:/path/path) causes any problems or incompatibilities. 
Paths still stay paths, we are only adding optional namespace prefixes to them, nobody is converting paths into windows style backslashes (\)

But lets say that namespaces were big showstopper we could always create a new method for accessing multi-backend systems like ```(getbv "vault" "/foo/morefoo")```  or ```(getbv "backend1" "/foo/morefoo")```

Anyway there seems to be many options to solve these problems, as long as confd community cares about having multiple backends capability. With all that said I can understand the argument of it being more hassle then it's worth and confd not wanting to go in such direction.

For my use case it was either write a simple custom system that behaves like confd or extend confd with multi backend capability. Obviously I chose the latter and figured someone else might also be in need of similar functionality.

odedlaz:
@fbrbovic, as far as I know @kelseyhightower doesn't care about backward compatibility until we reach the 1.0 milestone.

bacongobbler:
#414 is not something we're interested in supporting, so I'm closing this.

PR #419, opened 2016-03-25
add trimSuffix template function

dmexe:
Another simple strings function:
- **trimSuffix**


bacongobbler:
This is apparently a new feature in Go 1.6, so the 0.12-alpha3 release should already have it. See https://github.com/kelseyhightower/confd/issues/420 and https://golang.org/pkg/text/template/#hdr-Text_and_spaces for more information.

bacongobbler:
Oops! This is different. Re-opening.

Issue #402, opened 2016-02-26
Documentation for Installing from Source


jbkc85:
Though the documentation says just run ```go build```, the following packages are still telling me I need to use ```go get``` prior to the build.  Am I doing something wrong here or is this in fact the necessary steps?

```bash
[/repositories/git/opensource/confd]$ go build
config.go:14:2: cannot find package "github.com/BurntSushi/toml" in any of:
	/usr/local/go/src/github.com/BurntSushi/toml (from $GOROOT)
	/repositories/git/opensource/confd/src/github.com/BurntSushi/toml (from $GOPATH)
confd.go:10:2: cannot find package "github.com/kelseyhightower/confd/backends" in any of:
	/usr/local/go/src/github.com/kelseyhightower/confd/backends (from $GOROOT)
	/repositories/git/opensource/confd/src/github.com/kelseyhightower/confd/backends (from $GOPATH)
confd.go:11:2: cannot find package "github.com/kelseyhightower/confd/log" in any of:
	/usr/local/go/src/github.com/kelseyhightower/confd/log (from $GOROOT)
	/repositories/git/opensource/confd/src/github.com/kelseyhightower/confd/log (from $GOPATH)
confd.go:12:2: cannot find package "github.com/kelseyhightower/confd/resource/template" in any of:
	/usr/local/go/src/github.com/kelseyhightower/confd/resource/template (from $GOROOT)
	/repositories/git/opensource/confd/src/github.com/kelseyhightower/confd/resource/template (from $GOPATH)
```

```bash
[/repositories/git/opensource/confd]$ ./build
Building confd...
config.go:14:2: cannot find package "github.com/BurntSushi/toml" in any of:
	/usr/local/go/src/github.com/BurntSushi/toml (from $GOROOT)
	/repositories/git/opensource/confd/src/github.com/BurntSushi/toml (from $GOPATH)
confd.go:10:2: cannot find package "github.com/kelseyhightower/confd/backends" in any of:
	/usr/local/go/src/github.com/kelseyhightower/confd/backends (from $GOROOT)
	/repositories/git/opensource/confd/src/github.com/kelseyhightower/confd/backends (from $GOPATH)
confd.go:11:2: cannot find package "github.com/kelseyhightower/confd/log" in any of:
	/usr/local/go/src/github.com/kelseyhightower/confd/log (from $GOROOT)
	/repositories/git/opensource/confd/src/github.com/kelseyhightower/confd/log (from $GOPATH)
[...]

PR #431, opened 2016-04-12
README Adjustments

jbkc85:
README Adjustments per ticket https://github.com/kelseyhightower/confd/issues/402.  The noted changes are how I have had to install confd with the new vendor plugins, etc.  Please feel free to adjust where needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,17 +20,20 @@
 Go 1.6 is required to build confd, which uses the new vendor directory.
 
 ```
-git clone https://github.com/kelseyhightower/confd.git
-cd confd
-go build
+cd $GOPATH
+go get github.com/kelseyhightower/confd
+git clone https://github.com/kelseyhightower/confd
+go build github.com/kelseyhightower/confd
 ```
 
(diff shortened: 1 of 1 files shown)


bacongobbler:
Thanks! I adopted this and pushed some changes up to master in https://github.com/kelseyhightower/confd/commit/6386163046ddad465498d6e002bd8e5ab3748365. The reason why I don't want to recommend this approach is because `go get` will fetch the dependencies from their respective `master` branches and dump them all in your $GOPATH. Since we already have a `vendor/` directory with all our dependencies anyways, it's better to `git clone` and run `./build` instead, which is a simple script that runs `go build -o bin/confd .` sanely. Running `go get && go build/install` will use the dependencies fetched from master, which could be in a broken state in comparison to the vendored dependencies. 

Thank you very much for your help!

jbkc85:
@bacongobbler awesome, and thanks for the explanation!  I haven't used the vendor folder for any projects yet...so knowing the difference between a straight `git clone` via `go get` is extremely beneficial.

PR #445, opened 2016-05-10
add go version and git commit sha1 to version string

mountkin:
New features may be added to golang template.
If the user want to use the new feature (for example, go 1.6 supports trimming whitespace characters), they have to check the go version that the binary was built with. So I think adding the golang version to the `-version` command's output may be useful.

Signed-off-by: Shijiang Wei mountkin@gmail.com

### DIFF
--- a/build
+++ b/build
@@ -2,4 +2,10 @@
 
 echo "Building confd..."
 mkdir -p bin
-go build -o bin/confd .
+
+GIT_COMMIT=$(git rev-parse --short HEAD)
+if [ -n "$(git status --porcelain --untracked-files=no)" ]; then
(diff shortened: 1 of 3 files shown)


bacongobbler:
Personally, I'd want to create a release without the commit information. I don't think the commit is useful when we cut an official release, since most users would want to compare against a SHASUM instead. The runtime version is definitely a nice touch, though.

Leaving this up to @kelseyhightower

mountkin:
@bacongobbler The commit is useful if someone help us testing the binary built from master branch.

bacongobbler:
Yes, I agree it's useful for dev builds but I'm not sure it's useful to have it in official releases. Anyways, leaving this to Kelsey to decide.

PR #454, opened 2016-06-17
Adding check-only mode.

jsargiot:
This mode allows to process template resources and run check_cmd on the result without updating target files nor running reload_cmd. This mode is useful to run monitoring checks where it's nice to be able to detect problems with the templates at rendering and application level.

Example usage:

Confd is configured to update nginx.conf file and you need a way to detect problems with template. Right now `-noop` is ok to test template rendering but it won't catch problems in the nginx configuration (like duplicated upstreams), so this new command allows to render the template and run nginx template check.


bacongobbler:
Hey @jsargiot! So essentially this is just a dry run? Perhaps `--dry-run` would be a better name to call this flag to follow the convention of other projects. This sounds like a handy feature to support.

jsargiot:
> Perhaps --dry-run would be a better name to call this flag to follow the convention of other projects.

To me `--dry-run` means "pretend you are doing things" in this case the template is being rendered and check_cmd is actually executed... so it would feel a bit akward to me but I don't really have a strong opposition to it.

PR #457, opened 2016-06-20
Implemented Azure table storage backend

graemechristie:
Implemented an Azure table storage backend. I have implemented integration tests, but I have not enabled them as part of the build as they would require a live Azure Table Storage instance to run against (MS, in their infinite wisdom, have only release a local Table Storage service for Windows).


vendor/github.com/hashicorp/hcl/hcl/parser/test-fixtures/array_comment.hcl
```diff
@@ -1,4 +0,0 @@
-foo = [
```
graemechristie: Not sure why "godep" removed all of these ...

PR #397, opened 2016-02-23
yaml/json file backend

HeavyHorst:
I've createt a simple file based yaml/json backend. Closes #361 

It is working with json files:

```
{
    "MYAPP_DATABASE_URL": "db.example.com",
    "MYAPP_DATABASE_USER": "rob",
}
```

and yaml files:

```

---
  MYAPP_DATABASE_URL: "db.example.com"
  MYAPP_DATABASE_USER: "robby"
```

It has full watch support.
./confd -backend yaml -yaml-file test.yaml -watch

---

This change is [Reviewable](https://reviewable.io/reviews/kelseyhightower/confd/397)

PR #471, opened 2016-08-03
Json backend

ctchurch:
confd is useful for us to pull service config from etcd and vault, but we store dev-only config in Config.json files, so I added a json file backend.

If there are conventions or enhancements you'd like to see for this to come in, I'm happy to rework.


bacongobbler:
closing as a duplicate of #397

PR #481, opened 2016-09-08
Support metad backend

jolestar:
Add metad backend to confd.
[Metad](https://github.com/yunify/metad) is a metadata server tool, support self semantic. metad keep a mapping of IP and metadata, client direct request "/self", will get the metadata of current node.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,9 +37,9 @@ before_install:
   - mkdir /tmp/zookeeper && cp integration/zookeeper/zoo.cfg zookeeper-3.4.8/conf/zoo.cfg
   - zookeeper-3.4.8/bin/zkServer.sh start
   # Install metad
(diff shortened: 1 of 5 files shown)


kelseyhightower:
I really want to avoid adding more backend like this. The goal is to support K/V stores and avoid things like this. We really need a plugin system to avoiding ever increasing backends.

jolestar:
@kelseyhightower  What's the plan? Use a K/V stores lib like docker/libkv to replace backends or confd embed a K/V store? Our team plan to use confd (as client) and metad (as server)，to define a metadata standard for our third party developer, so we hope metad can be supported by confd.

bacongobbler:
I've gone ahead and made a v0.12.1 release on my fork that includes this change. See [the v0.12.0](https://github.com/bacongobbler/confd/releases/tag/v0.12.0) and [v0.12.1 release notes](https://github.com/bacongobbler/confd/releases/tag/v0.12.1).

PR #488, opened 2016-10-03
Upgrade Dockerfile to use latest stable Alpine which has Go 1.6

dmegyesi:
Hi everyone,

I wanted to use the "suppress newlines" feature introduced in Go 1.6 with confd. Turned out, the release binaries uploaded to Github are made with Go 1.5.
The latest, stable(!) Alpine version already ships with Go 1.6, so I thought we should upgrade the builder Dockerfile to use it.

The mentioned feature:

> {{end}} vs. {{- end}} and {{end -}} will suppress unnecessary empty lines in the confd generated output, creating a much human-readable output file.

### DIFF
--- a/Dockerfile.build.alpine
+++ b/Dockerfile.build.alpine
@@ -1,7 +1,6 @@
-FROM alpine:3.3
+FROM alpine:3.4
 
 ENV GOPATH /go
-ENV GO15VENDOREXPERIMENT 1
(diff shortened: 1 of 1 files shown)


bacongobbler:
Since this isn't a confd feature and rather a infra change, I can merge this. LGTM!

PR #496, opened 2016-10-26
Adding nested key support to Vault backend

bmitch3020:
This will recursively list keys in the vault backend so the conf.d/*.toml only needs to list the parent key. This still needs testing. I've included some untested updates to the vault integration tests, need more integration tests for the other backends, and I've made no changes to the unit tests.


bacongobbler:
Looks like travis is failing on this one:

> The command "bash integration/etcd/test.sh" exited with 1.
> The command "bash integration/vault/test.sh" exited with 1.
> The command "bash integration/zookeeper/test.sh" exited with 1.

Issue #504, opened 2016-11-15
confd work with redis, and how to set a specify database?


ouewqa:
I can't  find a example of the confd config file that work with redis.

who can give me a example?

how to set redis auth
how to set redis db
more...

bacongobbler:
Take a look at the [integration tests](https://github.com/kelseyhightower/confd/blob/master/integration/redis/test.sh) for redis. Let me know if you have any specific questions.

koolay:
Is it supported to connect special db?

bacongobbler:
I'm not sure what you mean by that. Can you provide a link or elaborate more on what you mean by "special db"?

koolay:
127.0.0.1:6379/8

HeavyHorst:
I looked at the redis code. Its not possible at the moment to specify a specific redis db.
But the redis client lib we are using supports it:
  - https://github.com/garyburd/redigo/blob/master/redis/conn.go#L112

To support this we may need to create a new flag. We could also add the database to the connection string (localhost:6379/4) for example.

HeavyHorst:
@koolay I have created a pull request (#505).
Feel free to test it !

PR #505, opened 2016-11-16
added support to specify a redis database (in the connection string)

HeavyHorst:
this addresses #504.

### DIFF
--- a/backends/redis/client.go
+++ b/backends/redis/client.go
@@ -2,10 +2,12 @@ package redis
 
 import (
 	"fmt"
-	"github.com/garyburd/redigo/redis"
 	"os"
+	"strconv"
 	"strings"
 	"time"
+
+	"github.com/garyburd/redigo/redis"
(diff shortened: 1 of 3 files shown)


bacongobbler:
Would you mind adding an additional test [here](https://github.com/kelseyhightower/confd/blob/master/integration/redis/test.sh#L23) with the following?

```
confd --onetime --log-level debug --confdir ./integration/confdir --interval 5 --backend redis --node 127.0.0.1:6379/0
```

Some docs [here](https://github.com/kelseyhightower/confd/blob/master/docs/quick-start-guide.md#redis-1) would also be really appreciated. That way we can test to ensure that both the old way and the new way work as intended, and can provide users like @koolay how to specify the database.

Asides from a few supporting changes, LGTM!

HeavyHorst:
@bacongobbler heyho
I've added the test and a small note in the quickstart guide.
Looks good in travis.

PR #517, opened 2016-12-22
No need loop

xusss:


okushchenko:
@xusss it's important to have this loop in place, or otherwise, confd will exit after the first iteration, making `WatchPrefix` unusable.
Compare:
1. With a loop: https://play.golang.org/p/uf94rfXkva
2. Without a loop: https://play.golang.org/p/MOmZFR9KLK

HeavyHorst:
@okushchenko the select blocks until one of its cases can run and every case returns.
The for loop is not needed.

okushchenko:
@HeavyHorst missed that function returns from inside select block, thanks.

PR #538, opened 2017-03-02
Adding nested key support to Vault backend v2

### LINKED ISSUE #494: Vault backend is not recursive
The vault backend calls `c.client.Logical().Read(key)` which will return items at that specific key, but not child keys. For that, I believe we need to also process the output of `c.client.Logical().List(key)`. Without the List, the conf.d/*.toml needs to identify each key you want to use in Vault and you can only iterate over multiple values of a single key, rather than multiple nested keys.

schuylr:
This is a fix to #496 where the test failure was due to a missing generic backend for `nested`.

I did not focus on the Zookeeper failure as it seems to be out of scope to the changes made.

It would be great to get this merged!

Fixes #494, closes #496


schuylr:
It looks like the failure now stems from trying to install DynamoDB - also considered out of scope for this PR.

schuylr:
I looked closer at the failures. The DynamoDB was related to AWS changing the download path to a full `.tar.gz` path, and the other ones were due to the nested values not being added to etcd and Zookeeper. All should be well now.

jstoja:
An idea why this is not merged ?

PR #541, opened 2017-03-03
feat: add option -env-sep which allow the env backend to use another char separator

nvanheuverzwijn:
Because the kelsey fork seems dead and you seems to be the community champion, here I am.
This PR is a shameless attempt to fix the problem reported in [here](https://groups.google.com/forum/#!topic/confd-users/k56ifrMmR28). It add a new option called -env-sep which allow the env backend to use another character separator. Actually, this character is _. This is not always practical because you cannot specify _ in a key while using env backend.

Regards,
Nic

### DIFF
--- a/backends/client.go
+++ b/backends/client.go
@@ -46,7 +46,7 @@ func New(config Config) (StoreClient, error) {
 	case "redis":
 		return redis.NewRedisClient(backendNodes, config.ClientKey)
 	case "env":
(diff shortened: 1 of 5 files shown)


nvanheuverzwijn:
Check this PR : https://github.com/kelseyhightower/confd/pull/567

okushchenko:
Closing in favor of #567

PR #564, opened 2017-05-29
Add elementary integer oprations and quorum.

ColinChartier:
We use confd for service discovery, and it is a huge pain to use a separate service to calculate n/2+1 for elasticsearch configuration.

This greatly simplifies configs without adding too much obscurity.


okushchenko:
Duplicate of #363

okushchenko:
Duplicate of #363

PR #567, opened 2017-06-02
Env backend support for keys with underscore

stela:
This patch allows underscores in paths, even when the "env" backend is used.

I attempted to add the key-value combinations for all the backend integration tests, but didn't actually run them besides for the "env" backend. The formats I saw didn't give me reason to suspect underscores aren't handled like other alphanumeric values in other backends, hope that's correct. Automated tests would have been great though.

### DIFF
--- a/backends/env/client.go
+++ b/backends/env/client.go
@@ -28,10 +28,12 @@ func (c *Client) GetValues(keys []string) (map[string]string, error) {
 	}
 	vars := make(map[string]string)
 	for _, key := range keys {
-		k := transform(key)
+		keyNoSlashPrefix := strings.TrimPrefix(key, "/")
+		k := transform(keyNoSlashPrefix)
 		for envKey, envValue := range envMap {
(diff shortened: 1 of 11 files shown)


nvanheuverzwijn:
This is a better solution than this https://github.com/kelseyhightower/confd/pull/541

Nice work

Danno040:
So, there's still a "gotcha" with this solution when using prefix for loading a group of keys.

For example, this will fail:

## basic.toml
```toml
[template]
mode = "0644"
src = "basic.conf.tmpl"
dest = "/tmp/confd-basic-test.conf"
keys = [
  "/database",
]
```

## basic.conf.tmpl
```
host={{getv "/database/host"}}
port={{getv "/database/user_name"}}
```

Because everything after the prefix gets "cleaned", the result of `GetValues` will look like:

```go
map[string]string{"/database/user/name":"something", "/database/host":"host"}
``` 

Maybe a minor problem, but worth pointing out.

PR #572, opened 2017-06-13
Change etcd node port because the official port of etcd is 2379

JohnHaan:
Change etcd node port because the official port of etcd is 2379


videte47:
👍 how can we get some traction on these PRs ?

okushchenko:
@JohnHaan, thanks

PR #601, opened 2017-09-11
Add StartCmd which runs when config does not exist yet

2rs2ts:
When confd creates a new file rather than updating an existing one, run
the StartCmd instead of the ReloadCmd. Useful for when you want to start
a service for the first time after writing its configs.

### DIFF
--- a/resource/template/resource.go
+++ b/resource/template/resource.go
@@ -47,6 +47,7 @@ type TemplateResource struct {
 	ReloadCmd     string `toml:"reload_cmd"`
 	Src           string
 	StageFile     *os.File
+	StartCmd      string `toml:"start_cmd"`
 	Uid           int
 	funcMap       map[string]interface{}
 	lastIndex     uint64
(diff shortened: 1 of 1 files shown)


2rs2ts:
Hey, not sure about these consul test failures. I'm not that familiar with the project, but it looks like the attempt to curl the locally running consul failed, which seems wholly unrelated to my changes.

okushchenko:
Usually, confd is used either in the entrypoint.sh script of a container, as a side car container or as ExecStartPre command in a systemd unit, making a StartCmd unnecessary. Can you describe your use case in more details?

2rs2ts:
@okushchenko We want to use the kubernetes API to detect when a new service that we want to load balance is deployed. On the load balancer machines, we use confd to create a haproxy config file and systemd unit for each of these services. It is convenient for us to create the new service with confd, then have confd enable and start it.

2rs2ts:
Closing due to lack of traction

Issue #532, opened 2017-02-16
redis backend hashes support 


agrakhov:
Hey guys, is it possible to add support for redis hashes?
They suggest it as best practice for memory optimization (https://redis.io/topics/memory-optimization), so would be nice to have it.

Thanks!

okushchenko:
Fixed in #604

PR #604, opened 2017-09-18
Redis Backend Updates: WatchPrefix Support, Hash Support, Standard Redis Key Naming Convention

jmooradi:
I've made some updates to the redis backend based on some needs for a project I was working on. I figured some of these might be useful for others as well as I wanted a second look at all the code for additional feedback on any problems you might see in the implementation especially since I'm fairly new to golang.

Overview of additions:
- WatchPrefix support using redis pubsub and keyspace events (if you don't want to enable keyspace events on your redis db you can also manually trigger a watchprefix using publish ex: `publish __keyspace@db__:prefix:key set`)
- GetValues rewrite to support redis hashes which is preferable since a normal scan has to iterate over every key in the entire database #532
- Convert confd "/" key naming notation to redis standard ":" notation when doing key lookups, this change will break existing setups, maybe it would be better to use a config flag to enable this?


backends/redis/client.go
```diff
@@ -110,49 +140,151 @@ func (c *Client) GetValues(keys []string) (map[string]string, error) {
...
+							return
+					}
+				}
+			}()
+
+			c.psc.PSubscribe("__keyspace@" + strconv.Itoa(db) + "__:" + transform(prefix) + "*")
+		}
+	}()
+
+	for {
```
okushchenko: This loop is redundant. The `WatchPrefix` function exists after a return.

backends/redis/client.go
```diff
@@ -11,17 +11,25 @@ import (
...
+	waitIndex uint64
+	err       error
+}
+
 // Client is a wrapper around the redis client
 type Client struct {
 	client   redis.Conn
 	machines []string
 	password string
+	psc redis.PubSubConn
```
okushchenko: Formatting.

jmooradi:
I added the changes you suggested and added a "separator" flag that I figured might be useful for other backends as well. -separator ":" enables the redis native notation, you can also specify any string really

Issue #607, opened 2017-09-27
error calling getv: key does not exist:/XXX/XXX


aixushuai:
env: confd-0.13.0  + etcdv3
Write multiple data to etcd at the same path, After the confd watch the event, the configuration is generated, his causes an error:key does not exist....
etcd data eg:
upstream/servers/app1/address 192.168.1.100
upstream/servers/app1/port 8080
upstream/servers/app1/weight 1
upstream/servers/app1/down false

I think this is the result of multiple events on etcd, causes no data to be obtained.
temporary solution: update backends/etcdv3/client.go. line:122
Add code: time.Sleep(time.Second * 2)

> 	for wresp := range rch {
			for _, ev := range wresp.Events {
				fmt.Println(string(ev.Kv.Key))
				// Only return if we have a key prefix we care about.
				// This is not an exact match on the key so there is a chance
				// we will still pickup on false positives. The net win here
				// is reducing the scope of keys that can trigger updates.
				for _, k := range keys {
					if strings.HasPrefix(string(ev.Kv.Key), k) {
						time.Sleep(time.Second * 2)
						return uint64(ev.Kv.Version), err
					}
				}
			}

okushchenko:
@aixushuai This issue is supposed to be fixed by #613. Please, reopen if the issue is still present.

PR #613, opened 2017-10-08
Replace godep with dep

### LINKED ISSUE #592: dns srv function is not perfect
Firstly, `-srv-domain` doesn't support **_etcd-client-ssl._tcp.example.com** which is etcd dns srv with ssl. Even if I use `-scheme https`

Secondly, when I use `-srv-record _etcd-client-ssl._tcp.example.com -scheme https` to replace, it shows 
```
2017-08-30T15:27:13+08:00 linux confd[9653]: INFO SRV record set to _etcd-client-ssl._tcp.example.com
2017-08-30T15:27:13+08:00 linux confd[9653]: INFO Backend set to etcd
2017-08-30T15:27:13+08:00 linux confd[9653]: INFO Starting confd
2017-08-30T15:27:13+08:00 linux confd[9653]: INFO Backend nodes set to 02.example.com:4758, 01.example.com:4758, 03.example.com:4758
2017-08-30T15:27:13+08:00 linux confd[9653]: ERROR client: etcd cluster is unavailable or misconfigured; error #0: unsupported protocol scheme "01.example.com"
; error #1: unsupported protocol scheme "03.example.com"
; error #2: unsupported protocol scheme "02.example.com"

2017-08-30T15:27:13+08:00 linux confd[9653]: FATAL client: etcd cluster is unavailable or misconfigured; error #0: unsupported protocol scheme "01.example.com"
; error #1: unsupported protocol scheme "03.example.com"
; error #2: unsupported protocol scheme "02.example.com"
```
It seems that `-scheme` doesn't work.

Finally, I use `-node` and it connects successfully. But it tells me that key doesn't exist, because confd prepends `/` to my key automatically, even if I set `prefix = ""` in toml.

okushchenko:
https://github.com/golang/dep is the new way to manage dependencies from Go team. It's much easier to use then godep. For example, godep makes it really hard to update a single dependency. In contrast, updating a dependency with dep is easy.  Also, dep binds packages to semantic versions instead of git commit hashes, so is much safer.

I've refactored build, test, install and clean binaries into a Makefile to provide a much cleaner interface without a bash shell.

Etcd package is updated, which hopefully fixes #592 #607


kelseyhightower:
Yes!!!

PR #630, opened 2017-11-10
fix(confd): etcd curl err when make integration

opera443399:
* output

```bash
~]$ make integration
(omitted)
curl: (6) Could not resolve host: value=10.0.1.11; Name or service not known
(omitted)
curl: (6) Could not resolve host: value=10.0.1.11; Name or service not known
(omitted)
```


okushchenko:
@opera443399 thanks!

Issue #629, opened 2017-11-09
confd cannot watch change from etcdv3 when change kv one by one quickly 


zyf0330:
If I change etcdv3 kv one by one quickly, then confd won't get second change by watching, so won't execute reload_cmd.

zyf0330:
My suggestion is, when watch a event, wait a few seconds and then start do changes. If there is new event, reset this timer.

okushchenko:
Sounds good. Can you try yourself at implementing it?

zyf0330:
Ok, I will try.

zyf0330:
Because of change about reusing previous watch, this problem is fixed at the same time. But this feature can avoid frequent reloading in other hand.

okushchenko:
Some people might actually want frequent reloading. Let's close this issue and have a discussion about whether the delay is actually needed in #634

PR #634, opened 2017-11-20
add delayed process for etcdv3 watch

zyf0330:
If there is no new change happening after 5 seconds since last change, the rest job of Processor will continue.


okushchenko:
@zyf0330 does it fix #629 ? Also, I'm not sure about the 5 seconds sleep, it's too much of a delay, I was thinking of something under 1 second, like 200ms or 400ms. What are you thoughts?

zyf0330:
This fix isn't needed anymore, because reusing previous `watch` has solved this.

zyf0330:
No, I just want another demand.
This feature can avoid frequently reloading when multiple keys are changed.

zyf0330:
Please tell me a situation where people want frequent reloading.

okushchenko:
@zyf0330 it's not about reloading frequency, but rather about feedback latency. For example, when confd is used to configure Nginx I want the change to happen as soon as possible.

zyf0330:
Ok, maybe we can set a flag to open it optionally.

zyf0330:
Just fix a bug. If `rch` is closed, `case wresp = <-rch:` will loop forever and cpu usage high.

Issue #606, opened 2017-09-26
Fatal error when missing parameters with SSM Parameter Store backend


gdw2:
I have a template that has defaults defined for each variable.  When I run confd with the 'env' backend, missing parameters are replaced with the defaults.  When I run with the ssm backend, I get the following fatal error:

```
2017-09-26T17:03:57Z eb75bd4829e5 confd[91]: FATAL ParameterNotFound: Parameter /prefix/test/host not found.
        status code: 400, request id: af3f5ecf-a2dc-11e7-a6ef-1d6619482473
```

I would expect no error and the parameters to be replaced with the defaults.

okushchenko:
Can you share this template?

gdw2:
Template:
```
[FOO]
host:       {{getv "/test/host"     "default"}}
```

Run with: `confd -onetime -prefix "prefix" -backend ssm`

PeteE:
Any chance of this getting fixed?

okushchenko:
I can only reproduce the error in case when there is no such prefix in the store at all, in which case it's expected to throw an error. Try extending the `keys` part of the confd config to a prefix with at least 1 key in it.
Should we change this behavior?

gdw2:
I created the PR (referenced above).

PR #639, opened 2017-12-11
AWS ssm - don't treat "ParameterNotFound" as a fatal error

gdw2:
This PR checks the SSM response error string for "ParameterNotFound" and, if found, ignores the error.  Had I been more proficient at golang, I might have checked something more substantial like a status code.   

This is in an effort to fix https://github.com/kelseyhightower/confd/issues/606.

### DIFF
--- a/backends/ssm/client.go
+++ b/backends/ssm/client.go
@@ -2,7 +2,7 @@ package ssm
 
 import (
 	"os"
-	"string"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
--- a/backends/ssm/client.go
+++ b/backends/ssm/client.go
@@ -2,6 +2,7 @@ package ssm
 
 import (
 	"os"
+	"string"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


okushchenko:
@gdw2 I've simplified the error checking. Please, take a look.

gdw2:
@okushchenko , beautiful.  I need to learn go.

BlackAR:
@okushchenko I was attempting to test this, and the strings.Contains() version of checking seemed to work, while the err.Error() version did not. Could this be because the error includes the name of the parameter it failed on, so will never equal ssm.ErrCodeParameterNotFound?

HeavyHorst:
ssm.ErrCodeParameterNotFound is of Type Error. This needs to be:
err != ssm.ErrCodeParameterNotFound

Anthony Black <notifications@github.com> schrieb am Mo., 18. Dez. 2017,
17:30:

HeavyHorst:
Oh it's not, sorry.

Rene Kaufmann <kaufmann.r@gmail.com> schrieb am Mo., 18. Dez. 2017, 17:49:

okushchenko:
It looks like aws-go-sdk has a special type for errors. Fixed in #647 
@BlackAR @HeavyHorst fyi

PR #649, opened 2017-12-28
Add kubernetes backend

jfrabaute:
This backend can handle only three different entities for now:

services
endpoints
statefulsets

The path should be in the form:

/<type>/{namespace}

the namespace is optional.

Examples:

/services/default : Will monitor all the services in 'default' namespace
/services: Will monitor all the services in all the namespaces


mingfang:
Is this dead?

xiaoqge:
If confd directly watch pod, don't listen to svc changes, so you don't have to go through svc ipvs nat, performance will be better?

xiaoqge:
@jfrabaute @okushchenko Is this dead?

okushchenko:
@mingfang @xiaoqge no, this PR is not dead. It will be merged right after the #568. If You want to help - give your feedback in #568 or add a review comment.

jfrabaute:
For info, this has been used in production for a long time right now without any problem.
I'd like to make it a plugin, so I will change it to a plugin when #568 is merged.

Regarding improvements, I might switch from the k8s official client to this one: https://github.com/ericchiang/k8s which is simpler, but not sure yet.

@xiaoqge I have not tried ipvs yet so I can't comment. I'll investigate on this. Thanks.

PR #653, opened 2018-01-05
remove fmt.Println

wrfly:
replace fmt.Println() with log.Debug()

### DIFF
--- a/backends/etcdv3/client.go
+++ b/backends/etcdv3/client.go
@@ -3,7 +3,6 @@ package etcdv3
 import (
 	"crypto/tls"
 	"crypto/x509"
-	"fmt"
 	"io/ioutil"
 	"strings"
 	"time"
@@ -11,6 +10,7 @@ import (
 	"golang.org/x/net/context"
 
 	"github.com/coreos/etcd/clientv3"
+	"github.com/kelseyhightower/confd/log"
 )
 
 // Client is a wrapper around the etcd client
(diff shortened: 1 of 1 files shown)


backends/etcdv3/client.go
```diff
@@ -119,7 +120,7 @@ func (c *Client) WatchPrefix(prefix string, keys []string, waitIndex uint64, sto
 
 	for wresp := range rch {
 		for _, ev := range wresp.Events {
-			fmt.Println(string(ev.Kv.Key))
+			log.Debug(fmt.Sprintf("Key updated %s", string(ev.Kv.Key)))
```
okushchenko: `Log.Debug` does Sprintf for you, so it's redundant here.
https://github.com/kelseyhightower/confd/blob/master/log/log.go#L51-L54

wrfly: Sorry about that, I'll amend it.

wrfly: @okushchenko done, PTAL

backends/etcdv3/client.go
```diff
@@ -119,7 +119,7 @@ func (c *Client) WatchPrefix(prefix string, keys []string, waitIndex uint64, sto
 
 	for wresp := range rch {
 		for _, ev := range wresp.Events {
-			fmt.Println(string(ev.Kv.Key))
+			log.Debugf("Key updated %s", string(ev.Kv.Key))
```
okushchenko: There is no such method as `Debugf`, please change in to `Debug`

wrfly: embarrassing...

PR #654, opened 2018-01-17
Feature/global slice

wrfly:
Add three functions:
- makeSlice
- appendSlice
- rangeSlice

The slice is global, inspired by <https://stackoverflow.com/questions/36521839/how-to-create-a-global-variable-and-change-in-multiple-places-in-golang-html-tem>

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -234,6 +234,29 @@ Returns the string representing the decoded base64 value.
 key: {{base64Decode "VmFsdWU="}}
 ```
 
+### makeSlice
+
+Returns the global slice ID.
+
+```
+{{$sliceID := makeSlice "sliceID"}}
+```
+
+### appendSlice
+
+Appends a string to the slice.
+
+```
+{{appendSlice $sliceID "Value"}}
+```
+
+### rangeSlice
+
+Returns the data in that global slice.
+```
+{{rangeSlice $sliceID}}
+```
+
 #### Add keys to etcd
 
 ```
--- a/resource/template/template_funcs.go
+++ b/resource/template/template_funcs.go
@@ -39,6 +39,9 @@ func newFuncMap() map[string]interface{} {
 	m["reverse"] = Reverse
 	m["sortByLength"] = SortByLength
 	m["sortKVByLength"] = SortKVByLength
+	m["appendSlice"] = AppendSlice
+	m["makeSlice"] = MakeSlice
+	m["rangeSlice"] = RangeSlice
 	m["add"] = func(a, b int) int { return a + b }
 	m["sub"] = func(a, b int) int { return a - b }
 	m["div"] = func(a, b int) int { return a / b }
@@ -210,4 +213,44 @@ func Base64Encode(data string) string {
 func Base64Decode(data string) (string, error) {
 	s, err := base64.StdEncoding.DecodeString(data)
 	return string(s), err
+}
+
+type sliceTyp struct {
+	id   string
+	data []string
+}
+
+var globalSlices = make(map[string]sliceTyp)
+
+func AppendSlice(toSliceID string, str string) string {
+	if _, exist := globalSlices[toSliceID]; !exist {
+		globalSlices[toSliceID] = sliceTyp{
+			id:   toSliceID,
+			data: []string{str},
+		}
+	} else {
+		globalSlices[toSliceID] = sliceTyp{
+			id:   toSliceID,
+			data: append(globalSlices[toSliceID].data, str),
+		}
+	}
+	return ""
+}
+
+func MakeSlice(id ...string) string {
+	var ID = time.Now().String()
+	for _, i := range id {
+		ID = i
+	}
+
+	globalSlices[ID] = sliceTyp{
+		id:   ID,
+		data: []string{},
+	}
+
+	return ID
+}
+
+func RangeSlice(sliceID string) []string {
+	return globalSlices[sliceID].data
 }
--- a/resource/template/template_test.go
+++ b/resource/template/template_test.go
@@ -684,7 +684,8 @@ key: Value
 		updateStore: func(tr *TemplateResource) {
 			tr.store.Set("/test/data", `VmFsdWU=`)
 		},
-	}, templateTest{
+	},
+	templateTest{
 		desc: "seq test",
 		toml: `
 [template]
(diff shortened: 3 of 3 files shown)


okushchenko:
Can you share a use case that is being covered by these extra functions?

wrfly:
Hello @okushchenko , actually I did face some problems that `confd` cannot handle by now.
For example, I have a `etcd v3` stores some data:

```
/nodes/online/<hostA>: 1.2.3.4
/nodes/online/<hostB>: 4.5.6.7

/nodes/info/<hostA>/ip : 1.2.3.4
/nodes/info/<hostA>/ports/http: 8080
/nodes/info/<hostA>/ports/grpc: 9090
/nodes/info/<hostB>/ip : 4.5.6.7
/nodes/info/<hostB>/ports/http: 8080
/nodes/info/<hostB>/ports/grpc: 9090
```

And I have an Nginx who wants to proxy these HTTP and grpc ports, the config file should be:

```
upstream http{
  server 1.2.3.4:8080;
  server 4.5.6.7:8080;
}
upstream grpc{
  server 1.2.3.4:9090;
  server 4.5.6.7:9090;
}

server{
  listen 10001;
  proxy_pass http;
}
server{
  listen 10002;
  proxy_pass grpc;
}
```

Without these new funcs, the template file need to range every nodes in each upstream:

```
upstream http{
{{ range gets "/nodes/online/*" -}}
    {{ $hostname := replace .Key "/nodes/online/" "" 1 -}}
    {{ $ip := getv (printf "/nodes/info/%s/ip" $hostname) -}}
    {{ $http := getv (printf "/nodes/info/%s/ports/http" $hostname) -}}
  server {{ printf "%s:%.f;" $ip $http }}
{{end}}
}

upstream grpc{
{{ range gets "/nodes/online/*" -}}
    {{ $hostname := replace .Key "/nodes/online/" "" 1 -}}
    {{ $ip := getv (printf "/nodes/info/%s/ip" $hostname) -}}
    {{ $grpc := getv (printf "/nodes/info/%s/ports/grpc" $hostname) -}}
  server {{ printf "%s:%.f;" $ip $grpc }}
{{end}}
}
```
`confd` have already ranged all the data at first time, it's no need to fetch the data again and again.

I have to admit that the data arranged is not good, it's my fault. But I still think it's good to make `confd` a little complexer than `text/template`.

okushchenko:
Confd does fetching of keys only once per template, so at the moment when a template is being rendered keys were already fetched. I'm against adding any more unnecessary functions or additional complexity without a strong use case behind it.

wrfly:
Yeah, you are right. I didn't know that `confd` feches keys only once and by adding more debug log, I realized it does.

Thus, it's no need to add this function only for "pretty template file".

I'm closing this PR. Thanks.

wrfly:
Hello @okushchenko , I thought I need to reopen this PR because I faced another problem and others may need it.

In my case, I didn't know whether is there a upstream server for:

```
upstream http {
  ???
}
```

And there is no key to check with `{{ if exists "/no/key/to/check" }}`, so the only way to check if there is any upstream server is to range all the keys and check the length of the data fetched:

```
{{ if ne (len (rangeSlice $http)) 0 -}}
upstream http {
{{ range (rangeSlice $http) -}}
    {{ printf "    " }}server {{ . }};
{{ end -}}
}
{{ end }}
```

Besides, better organization of the template is another benefits.

hubo1016:
@wrfly Surely guys in Google are smarter and they have invented a "with" statement
```
{{with pipeline}} T1 {{end}}
	If the value of the pipeline is empty, no output is generated;
	otherwise, dot is set to the value of the pipeline and T1 is
	executed.

{{with pipeline}} T1 {{else}} T0 {{end}}
	If the value of the pipeline is empty, dot is unaffected and T0
	is executed; otherwise, dot is set to the value of the pipeline
	and T1 is executed.
```

So it is
```
{{ with $http -}}
upstream http {
{{ range . -}}
    {{ printf "    " }}server {{ . }};
{{ end -}}
}
{{ end }}
```

wrfly:
@hubo1016 But what is `$http` since I don't know there is any hosts that provides HTTP service... Only `range` can get those keys.

hubo1016:
@wrfly 
```
{{ with gets "/nodes/online/*" -}}
upstream http {
{{ range . -}}
    {{ printf "    " }}server {{ . }};
{{ end -}}
}
{{ end }}
```

`range` is a statement. It is not a function that returns a slice. It is a loop identifier. `gets` returns a slice.

hubo1016:
Another example from my project which is more complicated
```
{{- with lsdir "/discovery/jdb.op.zodiac.storage"}}

storage:
  {{- range . }}
  -
    name: {{ base . }}
    max_file: 1000
    url: http://{{ index (ls (print "/discovery/jdb.op.zodiac.storage/" .)) 0 }}/
    {{- $conf_key := base . | print "/configuration/jdb.op.zodiac.storage/endpoint/" -}}
    {{- if exists (print $conf_key "/redirect_url") }}
    redirect_url: {{getv (print $conf_key "/redirect_url")}}
    {{- end }}
    {{- if exists (print $conf_key "/internal_url") }}
    internal_url: {{getv (print $conf_key "/internal_url")}}
    {{- end }}
    {{- if exists (print $conf_key "/accel-redirect_url") }}
    accel-redirect_url: {{getv (print $conf_key "/accel-redirect_url")}}
    {{- end }}
  {{ end }}
{{- end }}
```

This example uses `lsdir` to iterate into the first level, and `ls` for the second level. It also use `print` to concat keys to create complicated relations.

wrfly:
I'm sorry, I mean I can only `range` those keys to find out whether there is a host that provide an HTTP service. Since there maybe a host online that doesn't provide HTTP service but others.
BTW, `range` can get 'values' too:

```
{{range pipeline}} T1 {{end}}
	The value of the pipeline must be an array, slice, map, or channel.
	If the value of the pipeline has length zero, nothing is output;
	otherwise, dot is set to the successive elements of the array,
	slice, or map and T1 is executed. If the value is a map and the
	keys are of basic type with a defined order ("comparable"), the
	elements will be visited in sorted key order.
```

hubo1016:
@wrfly I see. You have to organize your keys in another way, to make it possible to filter keys with `lsdir` or `gets`. For example, put any HTTP service under `/nodes/info/*/ports/http` and use `with gets "/nodes/info/*/ports/http"`

wrfly:
You are right, the organization of keys have a little problem...
However since I have implemented that design, I need that feature. I can re-organize the keys in next release.

Thanks a lot~

PR #661, opened 2018-02-01
Vault TLS certificates auth method

dragoslav:
[Vault TLS Certificates Auth Method](https://www.vaultproject.io/docs/auth/cert.html)

Authentication type `-auth-type cert`.
If `-client-cert` and `-client-key` are not set or invalid, authentication will fail.


okushchenko:
Thanks for your contribution!

PR #665, opened 2018-02-08
backend/etcd: add scheme to srv discovered endpoints

mihu:
I've noticed that when using SRV discovery with etcd backend the etcd client complains about unknown protocol with endpoints being just `a.b.c:2379`. Prepending the scheme to the endpoints fixes the issue.


okushchenko:
Thanks for your contribution!

PR #668, opened 2018-02-16
Add atoi function to template funcs and test

### LINKED ISSUE #660: Counter-based loops with `seq`
I found the `seq` function listed in `template_funcs.go`. This may not be the best way to do this, but if it is, I'm coming up short.

I want to pass the number of log workers in this config as an env var (or other lookup) with a default of 1. The problem is that `seq` only takes `int` types and everything returned by `getv` is a string.
```
{{ $num_loggers := getv "/bro/node/numloggers" "1" }}
{{- range (seq 1 $num_loggers) }}
[logger-{{.}}]
type=logger
host=localhost
{{ end }}

```

I'm having a similar issue trying to compare another key using the `gt` function. My full template is in the [rocknsm/rock-containers](https://github.com/rocknsm/rock-containers/blob/master/bro/docker/bro_node.cfg.tmpl) repo.

Is there a way to cast the string to an int and failing if the cast fails? Is there a better way to handle this?

dcode:
This fixes #660. Included test case too.

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -356,6 +356,14 @@ Wrapper for [net.LookupIP](https://golang.org/pkg/net/#LookupIP) function. The w
 {{end}}
 ```
 
+### atoi
+
+Alias for the [strconv.Atoi](https://golang.org/pkg/strconv/#Atoi) function.
+
+```
+{{seq 1 (atoi (getv "/count"))}}
+```
+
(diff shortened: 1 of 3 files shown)


okushchenko:
Thanks for the contribution! I've added a doc string and will merge it soon.

PR #669, opened 2018-02-17
add parseBool to template_funcs

lucastheisen:
My team has found need to use values in template if:
```
{{ if getv "/environment/debug" }}
```
But since getv returns a string this turns into something much more annoying:
```
{{if or (eq (getv "/enable-development") "1") (eq (getv "/enable-development") "true")}}
```
_Or worse if you want to support more true/false values_.

So I just want to `strconv.ParseBool` to the `template_funcs`:
```
{{ if getv "/environment/debug" | parseBool }}
```


resource/template/template_funcs.go
```diff
@@ -211,3 +213,11 @@ func Base64Decode(data string) (string, error) {
 	s, err := base64.StdEncoding.DecodeString(data)
 	return string(s), err
 }
+
+func ParseBool(a string) bool {
+	val, err := strconv.ParseBool(a)
+	if err != nil {
```
okushchenko: I'm not sure if this is the right thing to do. It might be very unexpected for a user to get a `false` on error. Other functions, e.g. base64Decode return an error directly. I propose to make it consistent with the other functions and return error. This eliminates a wrapper function `ParseBool`, so you can write `m["parseBool"] = strconv.ParseBool`.

lucastheisen: @okushchenko , i understand your reluctance.  If only going by the principle of least surprise, it would make sense to have it behave _exactly_ like `strconv.ParseBool`.  And it would make the implementation simpler.  

In this instance, I was going for a _simpler template_.  If we just use a straight up pass through to `strconv.ParseBool` then we have to have an additional check to see if the value was defined before trying to use it (or provide a default).  So you end up with something like:
```
{{ if getv "/environment/debug" "false" | parseBool }}
```
instead of 
```
{{ if getv "/environment/debug" | parseBool }}
```
However, if it would help get it accepted, I would be happy to change it...  Wdyt?

okushchenko: The idea of templating in Confd follows the principle of *least surprise* over a *simpler template*, so let's keep it that way. But I can see your point, thanks for the explanation.

lucastheisen:
Per your request @okushchenko , I switched to straight pass through for `ParseBool`.

Issue #671, opened 2018-02-23
etcd watch can miss changes


trystanj:
This applies to, at least, etcd backends. I'm not familiar enough with other use-cases to comment on them - at a cursory glance it seems that some might be okay.

The core of the issue is that a watch triggers a fetch/update sequence, after which a watch is reestablished _at the "current" index_. If - for any reason - the fetch/update sequence fails, the error is simply logged via the `errChan`. The `lastIndex` has already been incremented [here](https://github.com/kelseyhightower/confd/blob/master/resource/template/processor.go#L103), but that **doesn't matter** because the new watch _always_ establishes with [index 0](https://github.com/kelseyhightower/confd/blob/master/backends/etcd/client.go#L127) - that is to say, "the current index, whatever that may be". So the change that was previously picked up and failed is never applied, and the new watch skips right past it.

I have actively observed this happening. It's easy to repro, too - just randomly return an error inside [template.process()](https://github.com/kelseyhightower/confd/blob/master/resource/template/resource.go#L348) and change your key a few times. confd will miss changes that error out.

There's actually another issue related to this. Any changes that happen between the first watch firing and the second watch being established are completely missed (again, due to the fact that the watch is established at "0"). There's a gap in time that isn't accounted for due to not using the correct index.

I hope I'm misunderstanding something, because it seems to make the watch very liable to miss changes. A tiny network blip when fetching from etcd, or a filesystem that takes too long to respond can cause it to get out of sync.

Let me know if I need to be more clear or provide examples!

hubo1016:
I'm creating a new implementation for etcdv3. It uses the etcd revision for waitIndex. The watch request is always processing in the background, and the latest revision seen from watch responses are stored. When [WatchPrefix](https://github.com/kelseyhightower/confd/blob/9b0ce8e6bb49eb46f27df33271e5edcc36a09746/backends/etcdv3/client.go#L95) is called, the latest revision is checked against the waitIndex to determine whether there are changes after last call. If the latest revision is not changed, the call waits until it is updated. Multiple resources with the same prefix shares the same watch.

When `process()` fails (template executing failed, check failed, reload failed), I think it should be OK to ignore this revision because we do not have other choices. We cannot afford to retry this revision until it succeeds because there may be errors in templates or stored values. The template will always be processed again when stored values are updated later.

For temporary failures in template processing like DNS lookup, maybe we can add an optional parameter to the template resource to allow it retry for some times, but it should be another issue.

[...]

PR #678, opened 2018-02-26
[In Progress]new etcdv3 implementation

hubo1016:
This should be a fix of #671

I'm not familiar with Golang and this is not carefully tested yet. This is for review & CI.

### DIFF
--- a/backends/etcdv3/client.go
+++ b/backends/etcdv3/client.go
@@ -4,18 +4,93 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"io/ioutil"
-	"strings"
 	"time"
 
 	"golang.org/x/net/context"
 
 	"github.com/coreos/etcd/clientv3"
-	"github.com/kelseyhightower/confd/log"
+	"sync"
 )
+
+// A watch only tells the latest revision
+type Watch struct {
+	// Last seen revision
+	revision int64
+	// A channel to wait, will be closed after revision changes
+	cond chan bool
+	// Use RWMutex to protect cond variable
+	rwl sync.RWMutex
+}
+
+// Wait until revision is greater than lastRevision
+func (w *Watch) WaitNext(ctx context.Context, lastRevision int64, notify chan<-int64) {
+	for {
+		w.rwl.RLock()
+		if w.revision > lastRevision {
+			w.rwl.RUnlock()
+			break
+		}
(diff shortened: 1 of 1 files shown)


hubo1016:
There is still a corner case when a connection failure happens between the watch response and the calling of `GetValues`. When this happens, the `GetValues` call fails and is not retried. If there are no further updates, the destination file is not updated even after the connection restore. It should be a rare case. We can add some retry logic in `GetValues` to prevent this from happening.

@okushchenko @trystanj What's your opinions?

hubo1016:
@okushchenko I think this is ready to merge... In fact I've already using this implementation in the production environment of my company for a month

hubo1016:
should also fix #693

EyciaZhou:
doesn't call `defer close(cancelRoutine)` is ok? because seems the goroutine below will leak also. 

	go func() {
		select {
		case <-stopChan:
			cancel()
		case <-cancelRoutine:
			return
		}
	}()

hubo1016:
Quite right I think. Should I remove the channel and wait for ctx.Done() instead?

hubo1016:
@EyciaZhou should be fixed

hubo1016:
Also fix #685

okushchenko:
@hubo1016 thanks for submitting a PR! Sorry for not getting back to you for so long. The code looks solid. The only thing that I'm getting a hard time to grasp is why have you decided to group the get requests using a transaction? Does it reduce the number of requests to Etcd? What was your reasoning behind that?

hubo1016:
@okushchenko this helps on the consistency of the configuration, ensuring all values are retrieved in the same revision. With the old implementation, if two changes are made in revision 1 and 2, both affect two keys (e.g. delete with prefix, and then insert two keys with transaction), confd may get the first key with value in revision 1, but the second in revision 2. Though finally the configuration will be synced to the latest revision, reloading an inconsistent configuration may cause errors.

It also reduces the latency of updating when there are a lot of keys.

tinder-andyphan:
@hubo1016 curious about this line 

`https://github.com/kelseyhightower/confd/pull/678/files#diff-8386a15dc01968620740d0c9083008c7R88`

and why it's necessary?  couldn't this revision possibly not exist due to the disconnect/error being, say, a new leader being reeelcted/repartitioned, not due to compaction?

PR #682, opened 2018-03-01
refine code and fix go vet ./...

chai2010:


chai2010:
i use libconfd now: https://github.com/openpitrix/libconfd

Issue #667, opened 2018-02-15
YML-file backend does not honor Keys list from toml file


sseide:
using the file backend with an yaml file the keys section from the toml file is ignored and ALL keys defined within the yaml file are exposed to the template. This produces wrong results for the templated file...

Example yaml backend file (test.yml):
<pre>
test:
  a:
    key: "value-a"
  b:
    key: "value-b"
</pre>

Template file (/etc/confd/templates/test.tmpl):
<pre>
My value is: {{ index (gets "/*/key") 0 }}
</pre>

Two different TOML files to either render key from "a" or "b":
<pre>
# toml file "test-a.toml" - should produce "My value is  {/a/key value-a}"
[template]
prefix = "/test"
src = "test.tmpl"
dest = "/tmp/test-a.txt"
keys = [
  "/a/"
]
</pre>
<pre>
# toml file "test-b.toml" - should produce "My value is  {/b/key value-b}"
[template]
prefix = "/test"
src = "test.tmpl"
dest = "/tmp/test-b.txt"
keys = [
  "/b/"
]
</pre>

Now running confd with these files produces the following output - look at output lines 11 and 12 - there should only be displayed the respective key "/a/key" or "/b/key" but not booth. This result differs from other backends like consul.

<pre>
/etc/confd/conf.d # confd -onetime -log-level debug -backend file -file /tmp/test.yml 
2018-02-15T11:38:52Z 23574631c64c confd[21]: INFO Backend set to file
2018-02-15T11:38:52Z 23574631c64c confd[21]: INFO Starting confd
2018-02-15T11:38:52Z 23574631c64c confd[21]: INFO Backend source(s) set to /tmp/test.yml
2018-02-15T11:38:52Z 23574631c64c confd[21]: DEBUG Loading template resources from confdir /etc/confd
2018-02-15T11:38:52Z 23574631c64c confd[21]: DEBUG Found template: /etc/confd/conf.d/template-a.toml
2018-02-15T11:38:52Z 23574631c64c confd[21]: DEBUG Loading template resource from /etc/confd/conf.d/template-a.toml
2018-02-15T11:38:52Z 23574631c64c confd[21]: DEBUG Found template: /etc/confd/conf.d/template-b.toml
2018-02-15T11:38:52Z 23574631c64c confd[21]: DEBUG Loading template resource from /etc/confd/conf.d/template-b.toml
2018-02-15T11:38:52Z 23574631c64c confd[21]: DEBUG Retrieving keys from store
[...]

PR #683, opened 2018-03-03
#667 fixed yaml backend

AndrewChubatiuk:


backends/file/client.go
```diff
@@ -34,33 +37,68 @@ func (c *Client) GetValues(keys []string) (map[string]string, error) {
...
-	nodeWalk(yamlMap, "", vars)
+	if len(keys) == 0 {
+		nodeWalk(yamlMap, "", vars)
+	} else {
+		for _, key := range keys {
+			filteredYamlMap, err := filterByPath(key, yamlMap)
+			if err != nil {
+				return vars, err
+			}
+			nodeWalk(filteredYamlMap, key, vars)
```
hubo1016: Why not use `nodeWalk` for all data first, and then filter the values from `vars`?

AndrewChubatiuk: yes, you're right, updated PR

PR #705, opened 2018-04-27
Adding support for multible kubernetes clusters in vault

AmrEleraky:


okushchenko:
Thanks for opening a PR! Closing this one in favor of #708 because that one is more generic (adds path support for other auth methods besides Kubernetes) and has the necessary integration tests.

PR #718, opened 2018-05-27
Client insecure mode

PascalChardon:
Added TLS client insecure mode for etcd(v2) backend to mimic "curl --insecure". Allow connections to secured etcd without cert verification.

### DIFF
--- a/backends/client.go
+++ b/backends/client.go
@@ -27,6 +27,7 @@ type StoreClient interface {
 
 // New is used to create a storage client based on our configuration.
 func New(config Config) (StoreClient, error) {
+
 	if config.Backend == "" {
 		config.Backend = "etcd"
 	}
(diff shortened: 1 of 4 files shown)


okushchenko:
@PascalChardon thanks for opening a PR!

I'm not 100% sure about adding a special environment variable just for that configuration option. I think that support for configuration from environment variables is something that confd lacks, but it needs a generic solution and not a one-off thing. Otherwise the PR looks good. I'll merge if you will remove the part that handles the environment variable.

PascalChardon:
Ok, You're the boss :-) I'll remove the part with env variable and submit again.

PascalChardon:
Hi Oleksandr,
I removed the code around env variable and had to fix a small issue on a peace of code I pulled with a merge. Unclear how this code can be in the master branch.

sreenandan:
Hi @okushchenko I am in need of this feature. I was able to make build and consume confd binary with -client-insecure
But, when is 0.17.0 going to be released?

PR #727, opened 2018-07-07
add file backend to quick start guide

abtreece:
Recent Tweets inspired a contribution of documentation. CI failed tangentially so I added a fix for that as well.


okushchenko:
@abtreece thank you for the contribution!

PR #732, opened 2018-07-11
Add a function for generate UUID

h17liner:
In some cases can be useful have a possibility to generate UUID during template rendering


okushchenko:
@h17liner thanks for opening a PR! It looks like a nice feature to have. The only concern that I have here is the fact that you create a custom function to generate UUID, which looks like a perfect job for an external library, like this one https://github.com/google/uuid. What do you think?

PR #735, opened 2018-07-16
Make sure dir is exist before create stage file.

CBike1103:
Make sure dir is exist before create stage file. This is to prevent the error that when the dir is not exist, the file will not be generated.


cam-smith-soe:
Is there anything preventing this being merged? It's the last item for the 0.17.0 milestone and looks like a simple & sane fix

mapperr:
I'm eagerly waiting for this release due to the [client insecure ](https://github.com/kelseyhightower/confd/pull/718) enhancement : )

PR #738, opened 2018-07-31
Add prepare run command

LTKH:
Good day.
I added support of edition running command.
In my opinion it will be usefull to make some additional actions before running reload.
For example I am planning to use it as downloader for template, stored on git.


okushchenko:
@LTKH thanks for opening a PR! This looks like a nice feature to have. Have you tried using it? I just want to make sure that it works the way you want before we merge it.

LTKH:
Yes, I have tested build with this feature. I am using confd with etcd. Function executed when app restarted or branch changed. Now it working on about 100 servers. This is exactly what I need.

PR #747, opened 2018-09-10
Added Token support for Consul

rafaelsierra:
Consul allows Tokens to be used when getting KVs, they are used to
provide granullar access to KVs (and improve security overall as your
Consul server will not be open to the world)


abtreece:
This looks good to me. Nice contribution @rafaelsierra.

PR #767, opened 2019-01-07
Replace Dep with New Go 11 Modules

msolimans:
Enabled Go 11 Modules; it is the new way to manage dependencies from Go team. much cleaner and easier than ever before. it allows to put the project anywhere without the need to have it inside $GOPATH! 

I've updated Makefile too, added `mod` instead and removed `dep` and also update `vendor` with all required packages that were resolved by go modules. 

For more info about new Go 11 Modules, https://github.com/golang/go/wiki/Modules


jbsmith7741:
@kelseyhightower any chance on getting this merged?

Issue #731, opened 2018-07-11
Vault PKI support?


azalio:
Hello!
Does confd support issuing certs and renewing procedure for them?

okushchenko:
No, it doesn't. Also, I don't see how it might even be a responsibility of the confd tool to manage the certificates. How do you think this feature might be added here?

azalio:
I'm going to use confd for ETCD and Vault values. And it might be convenient to use confd for vault PKI too because in another way I need to use one software for values from vault and another software for certs in vault. 
And from another point of view, certs just values for a common name (CN) as a key.

okushchenko:
Do you want these value to be exposed along the key/value objects from Vault or as a completely separate backend?

azalio:
I think a separate backend would be more useful than an integrated solution.

okushchenko:
I'll leave this issue open for now. If anyone is interested in implementing such backend I'll gladly review the PR.

tw3rp:
Created PR https://github.com/kelseyhightower/confd/pull/774 comments gladly welcome thanks!

CH-rahulprabhu:
Small write up https://rprabhu.dev/2019/vault-confd-automation/

tw3rp:
Small write up on how it works https://rprabhu.dev/2019/vault-confd-automation/

PR #774, opened 2019-02-18
[Storage-backend][vault pki] Add storage backend vault pki to template certs from vault

tw3rp:
This PR includes changes to add a new storage backend vault pki to template certs from vault

1. "Inherits" all the methods from the `vault` backend so authentication and all the flags remain the same
1. Certs are issues by listing keys as `/mountpath/rolename/commonname` eg. `/pki/issue/my-role/www.example.com` and the keys returned are `/pki/issue/my-role/www.example.com/certificate` and `/pki/issue/my-role/www.example.com/private_key`. This is so because it goes against confd design principles to pass in multiple arguments to a key
1. The cert and the key *must* be templated to the same file because it is an open issue to template to multiple files (https://github.com/kelseyhightower/confd/issues/256)
1. An example script is put in so we can split them into individual files once templated.

This is for issue https://github.com/kelseyhightower/confd/issues/731 to template certs as a new backend


CH-rahulprabhu:
Small write up https://rprabhu.dev/2019/vault-confd-automation/

tw3rp:
small write up on how it works https://rprabhu.dev/2019/vault-confd-automation/

PR #777, opened 2019-03-06
allows you to specify dynamodb endpoint.

xdhuxc:
the original code only allows to use the local dynamodb, but now, you can specify the dynamodb endpoint so that you can use the AWS's online Dynamodb services.


xdhuxc:
support the aws credentials profile, so that you can use many  access keys and secret keys at the same time in a credentials file, the only thing you need to do is that specify the parameter --profile when starting confd process.

xdhuxc:
repair the package error.

PR #780, opened 2019-03-28
add owner and group support

majakubiec:
This commit adds owner and group variable support and it should work as shown in quick-start-guide but it wasn't working before.
When you set `owner(group)` and `uid(gid)` variable is not set, it sets uid to a value of the `owner(group)` id.
When `uid(gid)` variable is set then `owner(group`) field is ignored


vedharish:
**Bump**

Have tested that this PR works fine.

PR #784, opened 2019-05-02
added AAC backend with watch support

sepiroth887:


sepiroth887:
This adds support for the [Azure App Configuration service](https://docs.microsoft.com/en-us/azure/azure-app-configuration/overview). 

A few new config flags and a "fake" (ticker based) watch implementation with jitter to avoid thundering herd. 

It's pretty basic and lacks some response code handling for 429 and 503

PR #785, opened 2019-05-07
Support nil FileInfo when error, since go 1.10

dmakhno:
Ref: https://github.com/golang/go/issues/26425

This helps if the monitored file was removed between discover and fileinfo.


dmakhno:
@okushchenko, how do I get pr reviewed and luckily merged?
ps. mentioning you, since last several prs were merged by your will and effort. :)

PR #790, opened 2019-06-12
Allow conf.d and templates directory to be symlinks

### LINKED ISSUE #741: confd failing to read conf.d and templates folders if they happen to be symbolic links
We believe that the flaw is here https://github.com/kelseyhightower/confd/blob/93c8904baa404add44565be4e42716eb391f83f9/util/util.go#L89.  From what we understand a `mode.IsDir()` for a symbolically linked folder would not return `true`

jsok:
`filepath.Walk()` ignores symlinks, so evaluate any symlinks as we recurse the root path.

Closes #741

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ before_install:
   - sudo mv vault /bin/
   - vault server -dev &
   # Install zookeeper
-  - wget http://www.eu.apache.org/dist/zookeeper/zookeeper-${ZOOKEEPER_VERSION}/zookeeper-${ZOOKEEPER_VERSION}.tar.gz
+  - wget https://archive.apache.org/dist/zookeeper/zookeeper-${ZOOKEEPER_VERSION}/zookeeper-${ZOOKEEPER_VERSION}.tar.gz
   - tar xzf zookeeper-${ZOOKEEPER_VERSION}.tar.gz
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -35,7 +35,7 @@ before_install:
   - sudo mv vault /bin/
   - vault server -dev &
   # Install zookeeper
-  - wget http://www.eu.apache.org/dist/zookeeper/zookeeper-${ZOOKEEPER_VERSION}/zookeeper-${ZOOKEEPER_VERSION}.tar.gz
+  - wget https://archive.apache.org/dist/zookeeper/zookeeper-${ZOOKEEPER_VERSION}/zookeeper-${ZOOKEEPER_VERSION}.tar.gz
```
jsok: Only the latest version (`3.4.14` as of right now) is available from /www.eu.apache.org.
Use the archive instead so URLs don't suddenly stop `404`'ing in CI.

jsok:
Is there a process for getting code reviewed? Can't seem to add any reviewers or labels.
@carmstrong @gabrtv @kelseyhightower @bacongobbler @okushchenko

bacongobbler:
The last time any code was merged to master was 11 months ago, and 0.16.0 was released over a year ago. Pretty sure the project isn't seeing any recent activity, and I haven't worked on confd since 2016. If you're interested in getting this merged, I'd suggest forking. :)

jsok:
@bacongobbler thanks for the response. It would probably manage a lot of expectations if the top level README indicated that this project is no longer maintained.

PR #807, opened 2019-11-07
Auto Discovery of keys in tmpl file

rita-g:
In the current confd version you must specify the keys in the toml file.
This is a repetition of the DRY principle, as they are already declared in the tmpl file.
When using SSM as a backend, you can specify the keys as = [""], but that will retrieve all the keys from SSM, which causes a major performance hit.

Instead of repeating the keys in the toml file, AutoDiscoverKeys flag allows to read the tmpl file and fetch only the relevant keys.

Add AutoDiscoverKeys flag that retrieves from Store only the referenced keys in the tmpl file.
Add SSM improvement to allow query by specific key names, instead of by path prefix.

### DIFF
--- a/backends/client.go
+++ b/backends/client.go
@@ -84,7 +84,7 @@ func New(config Config) (StoreClient, error) {
 		log.Info("DynamoDB table set to " + table)
 		return dynamodb.NewDynamoDBClient(table)
 	case "ssm":
-		return ssm.New()
+		return ssm.New(config.QueryMode)
 	}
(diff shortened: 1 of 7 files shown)


rita-g-walkme:
@abtreece @okushchenko @kelseyhightower can this please be reviewed?

abtreece:
I'll take a look @rita-g-walkme. If it gets merged it would likely be in my fork (abtreece/confd) as work on confd in this location is basically dead.

